PR #7, opened 2015-09-21
Changing handling of folders called 'plugins'

MSLaguana:
Remotebuild was too loose in determining which folder is the plugins folder. This change limits it to the plugins folder in the root, so folders under /www or other locations should be handled normally.


nisheetjain:
:+1:

PR #29, opened 2015-09-25
Fixed a telemetry bug, which caused telemetry code to crash during install-reqs

digeff:
We were using normal lambda functions instead of typescript functions, so the this pointer was undefined.
Changing to typescript lambda functions should fix the crash


nisheetjain:
I agree with Leo, we should try forcing telemetry to these paths via tests. Moreover I thought we agreed on not using lambda functions at some point (coding guidelines)?

digeff:
Are you referring to Arrow Function vs. Normal Javascript Function?
We have a lot of Javascript Functions in the codebase.

nisheetjain:
:+1:

digeff:
I've just tested it with the current code that it's failing due to an issue with the latest version of the Android SDK:
Before this fix we got: Uknown error.
After this fix we get:

====================================================================================================

Android SDK
Downloading...
Installing...
Updating environment variables...
Configuring (this may take several minutes)...
Picked up _JAVA_OPTIONS: -Xmx512M
Picked up _JAVA_OPTIONS: -Xmx512M
'jar_pathswt_path*' is not recognized as an internal or external command,
operable program or batch file.


====================================================================================================

There were some errors while installing the third-party dependencies. Please see above for details.

Please restart the command prompt to ensure that any newly installed software can be used correctly.

leotlee:
:+1:

PR #31, opened 2015-09-28
Updating https cipher list

MSLaguana:
This change enforces that we use TLS1.2, and also explicitly specifies the ciphers that we permit rather than using a node-version-dependent list.

This specific list is a reordering of the one in node v4.1.1.


src/remotebuild/lib/server.ts
```diff
@@ -238,10 +238,34 @@ class Server {
...
+                  "!MD5",
+                  "!PSK",
+                  "!SRP",
+                  "!CAMELLIA"];
                 var sslSettings = {
                     key: certStore.getKey(),
                     cert: certStore.getCert(),
                     ca: certStore.getCA(),
+                    ciphers: cipherList.join(":"),
+                    secureProtocol: "TLSv1_2_server_method",
```
digeff: According to this: https://nodejs.org/api/tls.html

secureProtocol: The SSL method to use, e.g. SSLv3_method to force SSL version 3. The possible values depend on your installation of OpenSSL and are defined in the constant SSL_METHODS.

What version of OpenSSL are we using? Is it always the same? So we know this value will always work? Or can the user change it somehow?

MSLaguana: Previously we were using whatever the defaults were. With this fix, we force it to be TLS1.2 (aka SSL3.3) which is the most secure option at the moment.

src/remotebuild/lib/server.ts
```diff
@@ -238,10 +238,34 @@ class Server {
                 return Q(certStore);
             }).
             then(function (certStore: HostSpecifics.ICertStore): https.Server {
+              // A slight reordering of the node v4.1.1 cipher list
+              var cipherList = ["ECDHE-ECDSA-AES256-GCM-SHA384",
```
nisheetjain: why is the re-ordering needed? can we use crypto.getCiphers() here instead?

nisheetjain:
:+1:  assuming this works for node 0.10+

src/remotebuild/lib/server.ts
```diff
@@ -51,6 +53,13 @@ class Server {
         var app = express();
         app.use(expressLogger("dev"));
         app.use(errorhandler());
+        if (conf.secure) {
+            app.use(helmet.hsts({
+                maxAge: 1000 * 60 * 60 * 24 * 365, // one year in milliseconds
```
digeff: What do you think of using a small library for these kind of things like https://www.npmjs.com/package/milliseconds ?

src/remotebuild/lib/server.ts
```diff
@@ -51,6 +53,13 @@ class Server {
...
+                includeSubdomains: false,
```
digeff: Can you please add comments explaining why did we choose to set this property to false? (Same for all other properties)

MSLaguana:
To enable ECDHE it looks like we need a different kind of key; for now this is good enough (and better than using the node 0.12.x defaults)

PR #34, opened 2015-09-28
Adding a test for cordovaWrapper handling errors

MSLaguana:


nisheetjain:
should we add a case where cordova launches a process underneath and crashes?

MSLaguana:
What do you mean by that? Do you mean testing for hangs / deadlocks? At the moment we don't handle that and we'll continue waiting until the promise chain continues on.

dlebu:
LGTM

PR #35, opened 2015-09-29
Adding a test for remote build device builds.

joshuaskelton:


src/taco-cli/test/utils/serverMock.ts
```diff
@@ -38,6 +40,12 @@ class ServerMock {
                         setTimeout(() => {
                             response.writeHead(data.statusCode, data.head);
                             response.write(data.response);
+
+                            if (data.sendMockFile) {
```
MSLaguana: If you are always going to send a zip file, I'd name this property sendMockZipFile. Alternately, I'd let this property be a string (or falsy to indicate not doing it) and treat it as a filename to pipe through the response.

MSLaguana:
Looks good to me.

PR #37, opened 2015-09-29
Build parallel

MSLaguana:
With this change we try to build/run/emulate multiple platforms in parallel rather than sequentially.
This requires that we update some test expectations. Since we can build multiple remotes at once, we separate the remote servers to different ports to make the test expectations easier to check, and we update the test expectations so it no longer assumes that one build completes before the other starts, and thus some outputs of the first build could appear in the other.


nisheetjain:
:+1:

PR #40, opened 2015-09-30
Consolidating platform selection code

MSLaguana:
This changes the "what platforms should I build and where" code to be the same whether you do "taco [build/run/emulate]" or "taco [build/run/emulate] ios windows android".


src/taco-cli/cli/utils/settings.ts
```diff
@@ -137,78 +137,77 @@ class Settings {
...
-
-            return { location: buildLocation, platform: platform };
-        });
-    }
-
     public static loadSettingsOrReturnEmpty(): Q.Promise<Settings.ISettings> {
         return Settings.loadSettings().fail(function (): Settings.ISettings { return { remotePlatforms: {} }; });
     }
 
+    public static parseRequestedPlatforms(options: commands.ICommandData): string[] {
```
nisheetjain: please add comments here. Hard to follow what are we try to do here. does this need to be public?

src/taco-cli/cli/utils/settings.ts
```diff
@@ -137,78 +137,77 @@ class Settings {
...
+            }).concat(localPlatforms.map(function (platform: string): Settings.IPlatformWithLocation {
+                return { location: Settings.BuildLocationType.Local, platform: platform };
+            }));
+        }).then((platforms: Settings.IPlatformWithLocation[]) => {
+            var requestedPlatforms = Settings.parseRequestedPlatforms(options);
+
+            if (requestedPlatforms.length > 0) {
+                // Filter down to user-requested platforms if appropriate
+                return requestedPlatforms.map((platform: string): Settings.IPlatformWithLocation => {
+                    if (!options.options["local"] && platforms.some((p: Settings.IPlatformWithLocation): boolean => p.platform === platform && p.location === Settings.BuildLocationType.Remote)) {
```
nisheetjain: can we break this too long line into multiple lines?

src/taco-cli/cli/utils/settings.ts
```diff
@@ -87,7 +87,7 @@ class Settings {
             Settings.determinePlatformsFromOptions(options)
         ]).spread<Settings.IPlatformWithLocation[]>(function (supportedPlatforms: CordovaHelper.IDictionary<any>, platforms: Settings.IPlatformWithLocation[]): Settings.IPlatformWithLocation[] {
             var filteredPlatforms = platforms.filter(function (platform: Settings.IPlatformWithLocation): boolean {
-                var supported = !supportedPlatforms || platform.platform in supportedPlatforms;
+                var supported = !supportedPlatforms || platform.platform in supportedPlatforms || platform.location === Settings.BuildLocationType.Remote;
```
nisheetjain: not sure if I follow this part. why don't we log this warning under determinePlatformsFromOptions? in determinePlatformsFromOptions we are anyways dealing with supported platforms right? so we can say read all remote + local platforms, show a warning if this is not one of the supported ones, attach remote/local and return back IplatformWithLocation[]

src/taco-cli/cli/utils/settings.ts
MSLaguana: It does need to be public because some telemetry code needs to call it, to work out what platforms the user asked for.

src/taco-cli/cli/utils/settings.ts
MSLaguana: In determinePlatformsFromOptions we may end up with non-supported platforms. For example, you can "taco platform add ios" on windows, and then "taco build" will pick it up as part of 'all installed platforms'. Similarly someone could explicitly state "taco build ios --local" on windows, where ios is not a supported platform, and determinePlatformsFromOptions will include it as a local platform.

I didn't want to collapse both the determinePlatform and determinePlatformsFromOptions into the one function, since telemetry wants to see what the user asked for, not just what we ended up doing.

src/taco-cli/cli/utils/settings.ts
MSLaguana: Done

src/taco-cli/cli/utils/settings.ts
MSLaguana: Correction: This is about filtering out non-platforms, not platforms that aren't supported on this host.

The user could specify "taco build sndroid" or "taco build foo" and we know for a fact that cordova will not build these non-platforms, so we are filtering them out here. It also protects against non-platform folders in /platforms

nisheetjain:
:+1:

PR #41, opened 2015-09-30
Refactoring iOS build function

MSLaguana:
This splits the iOS build function into a common core, and iOS specific components. We move some globals into class member variables, and gives a better base for moving to supporting other platforms. This also fixes some lingering tslint issues in iosBuild.ts


nisheetjain:
:+1:

PR #42, opened 2015-09-30
Fix minor install reqs bugs

guillaumejenkins:
-Refusing install-reqs prompt no longer gives an error, but instead exits normally with an appropriate message
-install-reqs now ignores ".DS_Store" files on Mac when checking if install destination is empty
-install-reqs on Mac OS will no longer export environment variables multiple times if they already exist and point to the correct location(s)
-The install-reqs prompt is now consistent with the other prompts in TACO
-The Android SDK configuration step will no longer update/install the following packages: Tools (it comes installed by default with the Android SDK zip file), Extra Android Support (not needed to build projects), Extra Android M2 Repository (not needed to build projects)
-install-reqs summary now gives a rough estimation of download and disk size for the dependencies


src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -527,7 +532,7 @@ module TacoDependencyInstaller {
                 args = [];
             } else {
                 command = "sudo";
-                args = ["node"];
+                args = ["-E", "node"];
```
nisheetjain: need a comment what is -E for

src/taco-dependency-installer/installerRunner.ts
```diff
@@ -122,8 +122,13 @@ class InstallerRunner {
 
                 // Verify that the destination folder is empty if it already exists
                 if (fs.existsSync(value.installDestination)) {
-                    if (fs.readdirSync(value.installDestination).length > 0) {
-                        throw new Error(resources.getString("PathNotEmpty", value.displayName, value.installDestination));
+                    var dirItems: string[] = fs.readdirSync(value.installDestination);
+
+                    if (dirItems.length > 0) {
```
nisheetjain: how about we look for android.bat existence to conclude this?

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -122,31 +122,69 @@ class AndroidSdkInstaller extends InstallerBase {
 
         // Initialize values
         var androidHomeValue: string = path.join(this.installDestination, "android-sdk-macosx");
-        var addToPathTools: string = "$" + AndroidSdkInstaller.AndroidHomeName + "/tools/";
-        var addToPathPlatformTools: string = "$" + AndroidSdkInstaller.AndroidHomeName + "/platform-tools/";
-        var newPath: string = "\"$PATH:" + addToPathTools + ":" + addToPathPlatformTools + "\"";
-        var appendToBashProfile: string = "\n# Android SDK\nexport ANDROID_HOME=" + androidHomeValue + "\nexport PATH=" + newPath;
+        var fullPathTools: string = path.join(androidHomeValue, "tools/");
+        var fullPathPlatformTools: string = path.join(androidHomeValue, "platform-tools/");
+        var shortPathTools: string = "$" + AndroidSdkInstaller.AndroidHomeName + "/tools/";
```
nisheetjain: well you could be using ~/ notation?

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -122,31 +122,69 @@ class AndroidSdkInstaller extends InstallerBase {
...
+        var exportAndroidHomeLine: string = "";
+        var updateCommand: string = "";
+        var useShortPaths: boolean = true;
+
+        // Save Android home value
+        this.androidHomeValue = androidHomeValue;
+
+        // Check if we need to add an ANDROID_HOME value
+        if (!process.env[AndroidSdkInstaller.AndroidHomeName]) {
+            exportAndroidHomeLine = os.EOL + "export " + AndroidSdkInstaller.AndroidHomeName + "=\"" + androidHomeValue + "\"";
```
nisheetjain: we should use util.format here

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -122,31 +122,69 @@ class AndroidSdkInstaller extends InstallerBase {
...
+        }
+
+        // Check if we need to update .bash_profile
         var bashProfilePath: string = path.join(process.env.HOME, ".bash_profile");
-        var updateCommand: string = "echo '" + appendToBashProfile + "' >> '" + bashProfilePath + "'";
         var mustChown: boolean = !fs.existsSync(bashProfilePath);
 
-        this.androidHomeValue = androidHomeValue;
+        if (exportAndroidHomeLine || exportPathLine) {
+            updateCommand = "echo '" + os.EOL + "# Android SDK" + exportAndroidHomeLine + exportPathLine + "' >> '" + bashProfilePath + "'";
```
nisheetjain: util.format?

src/taco-dependency-installer/installerRunner.ts
guillaumejenkins: I was planning on implementing the "Patch Up" feature/suggestion bug in a separate changeset. This is just to make sure we don't crash / throw error when folder only contains a .DS_Store

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: Done

src/taco-dependency-installer/installers/androidSdkInstaller.ts
guillaumejenkins: Done

src/taco-dependency-installer/installers/androidSdkInstaller.ts
guillaumejenkins: Done

src/taco-dependency-installer/installers/androidSdkInstaller.ts
guillaumejenkins: I updated the expandEnvVariables helper method to look for a ~ at the start of the path and expand it

src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -108,9 +108,14 @@ module TacoDependencyInstaller {
 
                 telemetry.step("promptUserBeforeInstall");
                 return this.promptUserBeforeInstall()
-                    .then(function (): Q.Promise<number> {
-                        telemetry.step("spawnElevatedInstaller");
-                        return self.spawnElevatedInstaller();
+                    .then(function (acceptedPrompt: boolean): Q.Promise<number> {
```
digeff: Can you change this to an arrow function?

src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -317,9 +322,10 @@ module TacoDependencyInstaller {
             });
         }
 
-        private promptUserBeforeInstall(): Q.Promise<any> {
+        private promptUserBeforeInstall(): Q.Promise<boolean> {
             this.buildInstallConfigFile();
 
+            var self = this;
```
digeff: Please, use an arrow function instead...

src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -355,14 +372,14 @@ module TacoDependencyInstaller {
 
             return installerUtils.promptUser(resources.getString("YesExampleString"))
                 .then(function (answer: string): Q.Promise<any> {
-                    if (answer === resources.getString("YesString")) {
+                    if (answer.toLocaleLowerCase() === resources.getString("YesString")) {
```
digeff: Telemetry is using this library for this kind of y/n: https://www.npmjs.com/package/readline-sync#utility_methods-keyinynstrict I think it'd be great if we could change all the code to use that same library to unify the whole source code.

src/taco-dependency-installer/elevatedInstallerProtocol.ts
```diff
@@ -23,6 +23,7 @@ module ElevatedInstallerProtocol {
         CompletedWithErrors = 12,
         NoAdminRights = 33,
         CouldNotConnect = 37,
+        RefusedPrompt = 42,
```
digeff: Can you put in the comments a link on how you figure out which "code numbers" are unused? Also, do we have any way of knowing if either powershell or node decide in the future to start using any of these numbers?

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: That looks overkill for just checking for a "y" or a "Y"... Also, I believe we are going to go away from single key input (that is, the user will need to press "enter" in order for the input to be recorded) for the telemetry question. That library does not allow that behavior for a y/n prompt, so we would need to use just "question" method, which means we would need to manually check the answer with a check similar to what I'm doing here. There's also the fact that this is not localizable (the library doesn't let you customize the "y" and the "n" keys to something like "o" and "n" in French). I don't know if we're ever going to localize TACO, but it's something to keep in mind.

All in all, I don't think we should be taking an extra dependency for such a simple check.

digeff: That makes sense! What do you think about creating our own "utlity" method to ask for [y/n] so all the modules can use that same function?

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: Our earlier guideline was to use the "var self = this". A quick search in the code base shows that we still use the "self" notation in at least 70 other places.

If we decide we want to completely remove any "self" notation from our code, I would be inclined to add that to tslint and do the necessary changes in a separate changeset... What do you think?

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: See my other comment about our code conventions

src/taco-dependency-installer/elevatedInstallerProtocol.ts
guillaumejenkins: Don't think there is a way to predict the future for that =/

I'll add links to the pages where I got my information from

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: Yep, I think that would be the best solution... I think that it should probably be part of the same changeset that will modify the telemetry prompt though, so for now I will just make this quick change to at least make the prompt similar to what we have elsewhere

digeff: Ok! Please create a bug/issue/task so we can track that change!

src/taco-dependency-installer/installerRunner.ts
```diff
@@ -122,8 +122,13 @@ class InstallerRunner {
...
+                        // For Darwin platform, the directory can contain a ".DS_Store" file, which we need to ignore (the directory is essentially empty in that case)
+                        if (process.platform !== "darwin" || dirItems.length !== 1 || dirItems[0] !== ".DS_Store") {
```
meenakb: You might want to test if there is a remnant "thumbs.db" file upon android-sdk uninstallation on Windows.

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -36,11 +37,9 @@ import utilHelper = tacoUtils.UtilHelper;
 class AndroidSdkInstaller extends InstallerBase {
     private static ANDROID_HOME_NAME: string = "ANDROID_HOME";
     private static androidCommand = os.platform() === "win32" ? "android.bat" : "android";
-    private static ANDROID_PACKAGES: string[] = [
-        "tools",
+    private static ANDROID_PACKAGES: string[] = [    // IDs of Android Packages to install. To get the list of available packages, run "android list sdk -u -a -e"
+        // "tools",  // Android SDK comes by default with the tools package, so there is no need to update it. In the future, if we feel we want dependency installer to always install the latest Tools package, then uncomment this.
```
meenakb: Remove the additional "//" in the comment line

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -36,11 +37,9 @@ import utilHelper = tacoUtils.UtilHelper;
...
         "platform-tools",
-        "extra-android-support",
-        "extra-android-m2repository",
         "build-tools-19.1.0",
         "build-tools-21.1.2",
         "build-tools-22.0.1",
```
dlebu: Is there any way to fetch these programmatically and not having to do a release of taco every time a new Android version comes along?

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -122,31 +121,76 @@ class AndroidSdkInstaller extends InstallerBase {
...
+            if (existingSdkHome !== androidHomeValue) {
+                // A conflicting ANDROID_HOME already exists, warn the user, but don't add our own ANDROID_HOME
+                this.logger.logWarning(resources.getString("SystemVariableExistsDarwin", AndroidSdkInstaller.ANDROID_HOME_NAME, this.androidHomeValue));
+                useShortPaths = false;
+            }
+        }
+
+        // Check if we need to update PATH
+        if (!installerUtils.pathContains(fullPathTools)) {
+            addToPath += path.delimiter + (useShortPaths ? shortPathTools : fullPathTools);
```
dlebu: Shorter version:
```typescript
addToPath += path.delimiter + (shortPathTools || fullPathTools); 
```

src/taco-dependency-installer/installers/androidSdkInstaller.ts
guillaumejenkins: Being more intelligent about which packages we install will be in a separate changeset, we have a task for that :) For now I just wanted to remove these 2 packages, since they are not used to build Android projects

src/taco-utils/utilHelper.ts
```diff
@@ -219,6 +219,11 @@ module TacoUtility {
         public static expandEnvironmentVariables(str: string): string {
             var regex: RegExp = process.platform === "win32" ? /%(.+?)%/g : /\$(.+?)(?:\/|$)/g;
 
+            // For Mac OS, first try to detect the ~ notation at the start of the string and replace it with the HOME value
+            if (os.platform() === "darwin" && process.env.HOME) {
```
dlebu: I see a lot of if clauses in the code based on the os.platform(). It would be nice if we could have different implementations for the utility classes that use this a lot. installerUtils, utilHelper are good examples.

dlebu:
Overall, LGTM. What kind of test coverage do we have for these changes?

src/taco-dependency-installer/resources/en/resources.json
```diff
@@ -11,6 +11,10 @@
     "_InstallDestination.comment": "Label and value for the installation directory of a dependency. {0} is the path where the dependency will be installed.",
     "DependencyLicense": "License terms: <link>{0}</link>",
     "_DependencyLicense.comment": "Label and value for the url to the license terms of a dependency. {0} is the url.",
+    "DownloadSize": "Download size: {0} MB (approximately)",
```
nisheetjain: would recommend keeping size information in json file with units. so that we can show based on size, say 12 KB or 23 MB or 1 GB. this information doesn't need to be localized anyways

nisheetjain: never mind. we may need actual numbers to do math...

nisheetjain:
:+1:

PR #44, opened 2015-10-01
Tslint2

nisheetjain:
Enabling following tslint rules:
forin, label-position, no-conditional-assignment, no-empty, no-var-requires, quotemark, radix, use-strict


src/gulpmain.ts
```diff
@@ -21,7 +21,11 @@ import gulpUtils = require ("../tools/GulpUtils");
 /* tslint:disable:no-console */
 // Disable console rule for gulp file, since this is a build file
 // we don't want to take dependency on Logger here
+
+/* tslint:disable:no-var-requires */
+// var require needed to require package json
```
MSLaguana: This comment looks outdated; should refer to build_config.json?

src/taco-cli/cli/logo.ts
```diff
@@ -8,7 +8,11 @@
 // NOTE: This file is intended to run on-require, and print the logo, only once.
 // We avoid using logger.log and other package dependencies because this is required in
 // a couple of places including post-install, and we don't want to drag in all the infrastructure
+
+/* tslint:disable:no-var-requires */
+// var require needed to require package json
```
MSLaguana: An alternative is to just go with JSON.parse(fs.readFileSync()) instead of require(); not sure which I prefer more.

src/taco-cli/cli/utils/checkForNewerVersion.ts
```diff
@@ -56,7 +56,7 @@ class CheckForNewerVersion {
 
     public showOnExitAndIgnoreFailures(): void {
         try {
-            this.showOnExit().catch(error => { }).done();
+            this.showOnExit().catch(error => tacoUtility.UtilHelper.emptyMethod(error)).done();
```
MSLaguana: Why not just `this.showOnExit().catch(tacoUtility.UtilHelper.emptyMethod).done();` instead?

src/taco-cli/test/meta.ts
```diff
@@ -114,7 +114,7 @@ describe("taco meta command tests: ", function (): void {
                 Taco.runWithArgs(optionString).then(function (): void {
                     done(new Error("Passing Invalid options to \'taco\' should have failed"));
                 }, function (err: any): void {
-                    if (err.code = TacoErrorCodes.CordovaCommandFailed) {
```
MSLaguana: Good thing we enabled the check!

src/taco-cli/test/templates.ts
```diff
@@ -13,14 +13,21 @@
...
+/* tslint:disable:no-var-requires */
+// Special case to allow using color package with index signature for style rules
+var colors = require("colors/safe");
+/* tslint:enable:no-var-requires */
 
 import tacoUtils = require ("taco-utils");
 import Templates = require ("../cli/templates");
 import ms = require ("./utils/memoryStream");
 
-var colors = require("colors/safe");
```
MSLaguana: Actually I think this can already be an import; I just checked the colors.d.ts and they declare "colors/safe" as a module so it should pick it up fine.

src/taco-utils/helpCommandBase.ts
```diff
@@ -95,9 +95,10 @@ module TacoUtility {
             Logger.log(util.format("   <synopsis>%s %s</synopsis><br/>", this.cliName, "<COMMAND>"));
 
             var nameDescriptionPairs: INameDescription[] = new Array();
-            for (var i in this.commandsFactory.listings) {
-                nameDescriptionPairs.push({ name: i, description: this.commandsFactory.listings[i].description, category: this.commandsFactory.listings[i].categoryTitle });
-            }
+            var listings: any = this.commandsFactory.listings;
+            Object.keys(this.commandsFactory.listings).forEach(function(i: string) {
```
MSLaguana: you can replace this.commandsFactory.listings here with listings since you already named it.

src/taco-utils/testUtils/testHelper.ts
```diff
@@ -11,7 +11,11 @@
 
 "use strict";
 
-import should = require ("should");
```
MSLaguana: This one can stay as an import: since we directly refer to it in the rest of the file, the typescript compiler won't decide that it is unused.

src/taco-utils/utilHelper.ts
```diff
@@ -354,6 +354,16 @@ module TacoUtility {
 
             return args;
         }
+
+        /* tslint:disable:no-empty */
+        /**
+         * An explicit helper empty method, which can be used in scenarios like
+         * silent callbacks, catch all exceptions do nothing etc.
+         */
+        public static emptyMethod(args?: any): any {
```
MSLaguana: I'd go with a typing of `public static emptyMethod(args: any...): void` since all of these empty functions must be returning void, and we should allow for multiple function arities.

MSLaguana:
Looks good to me once the above comments are addressed!

src/taco-cli/cli/logo.ts
nisheetjain: actually I like this. considering we have only 3 of these in product code.

src/taco-cli/cli/utils/checkForNewerVersion.ts
nisheetjain: good catch.

src/taco-cli/test/templates.ts
nisheetjain: actually that gives colors.strip as not defined which is being used by the test. I checked the latest type definition and that also doesn't define this..

src/taco-utils/helpCommandBase.ts
nisheetjain: yes done

src/taco-utils/testUtils/testHelper.ts
nisheetjain: done

src/taco-utils/utilHelper.ts
nisheetjain: sure

PR #45, opened 2015-10-01
Initial implementation of 'taco simulate' command.

TimBarham:
(New PR from branch in Microsoft/TACO -  same code as before)

This is still a work-in-progress (I'm not intending to commit this as-is), but I wanted to get feedback before I progress further. Things still to do:

* Finalize command line options
* Finalize help text
* Switch to using taco-simulate module once it is available
* Support telemetry
* Anything else?

Also, per suggestion from Jimmy - I'll modify `cordova-simulate` (or `taco-simulate` once that is a thing) to return a promise from the `simulate()` call.


src/taco-cli/cli/simulate.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ ? *******************************************************
```
MSLaguana: Looks like you might have been bitten by the BOM; at least I assume that's why there's a bunch of question marks here.

TimBarham: Hmmm.... I thought I had fixed that. Sigh.

src/taco-cli/cli/commands.json
```diff
@@ -497,6 +497,24 @@
...
+            "args": [
+                {
+                    "name": "PLATFORM", "description": "[CommandSimulatePlatformDescription]"
+                }
+            ],
+            "options": [
+                {
+                    "name": "--target", "description": "[CommandSimulateTargetDescription]"
+                }
+            ]
```
nisheetjain: if useful, you can add examples here. for e.g. look at examples for kit command

TimBarham: Done. May change after Soak reviews, but at least provides some placeholder options.

TimBarham:
I'm going to close this and open a fresh PR.

PR #47, opened 2015-10-02
Updating streamToPromise to handle 'end' events

MSLaguana:
The gulp-tslint stream only emits an 'end' event, not a 'finished' event, so the streamToPromise function wasn't noticing the end and things ended in a deadlock.


nisheetjain:
:+1:

PR #50, opened 2015-10-02
Batching cordova.raw.X calls and refactoring

MSLaguana:
It turns out that if you make concurrent calls to cordova.raw.build() with different platforms it mixes the platforms together, so calling build(["ios"]); build(["android"]) will try to build both ios and android in both invocations.

With this change we delegate all local build ordering concerns to cordova by making one invocation with the full list of local platforms.

I've also factored out some common code to clean things up a bit and help future code changes in this area.


src/taco-cli/cli/utils/settings.ts
```diff
@@ -102,6 +102,30 @@ class Settings {
         });
     }
 
+    /**
+     * Apply functions to all local and remote platforms.
+     * Local platforms are batched into one list because Cordova does not cope with multiple concurrent invocations
+     * Remote platforms are done individually (but concurrently with each other and local platforms) since they may
+     * go to completely different servers.
+     */
+    public static operateOnPlatforms(platforms: Settings.IPlatformWithLocation[],
```
dlebu: Is the Settings class the right place for this method? This looks more like a utility method.

src/taco-cli/cli/build.ts
```diff
@@ -146,26 +146,19 @@ class Build extends commands.TacoCommandBase {
...
-                            // Just build remote, and failures are failures
-                            return Build.buildRemotePlatform(platform.platform, commandData, telemetryProperties);
-                        default:
-                            return Q.reject(errorHelper.get(TacoErrorCodes.CommandBuildInvalidPlatformLocation, platform.platform));
-                    }
-                });
-            }));
+            var cleanPromise = Q({});
+            if (commandData.options["clean"]) {
+                cleanPromise = Q.all(platforms.map((platform: Settings.IPlatformWithLocation) => {
```
digeff: Do we have any code guidelines of:
() => { return X; } vs () => X;? I personally find the 2nd option easier to read. What do you think?

src/taco-cli/cli/build.ts
```diff
@@ -146,26 +146,19 @@ class Build extends commands.TacoCommandBase {
...
-                        case Settings.BuildLocationType.Remote:
-                            // Just build remote, and failures are failures
-                            return Build.buildRemotePlatform(platform.platform, commandData, telemetryProperties);
-                        default:
-                            return Q.reject(errorHelper.get(TacoErrorCodes.CommandBuildInvalidPlatformLocation, platform.platform));
-                    }
-                });
-            }));
+            var cleanPromise = Q({});
+            if (commandData.options["clean"]) {
```
digeff: I'd personally write this with an inline if, but the variable version is also fine:
commandData.options["clean"]
  ? Q.all(platforms.map((platform: Settings.IPlatformWithLocation) => Build.cleanPlatform(platform, commandData)))
  : Q({})

src/taco-cli/cli/utils/settings.ts
MSLaguana: I was trying to work out where best to put this. It is a utility method, but one that works specifically with Settings.IPlatformWithLocation,

src/taco-cli/cli/build.ts
MSLaguana: I split it into two lines because it was getting long, and if you do that then you need to add the braces or it complained.

src/taco-cli/cli/run.ts
```diff
@@ -160,17 +160,12 @@ class Run extends commands.TacoCommandBase {
...
-                        // Just run local, and failures are failures
-                        return CordovaWrapper.run(commandData, platform.platform);
-                    case Settings.BuildLocationType.Remote:
-                        // Just run remote, and failures are failures
-                        return Run.runRemotePlatform(platform.platform, commandData, telemetryProperties);
-                }
-            }));
+                buildTelemetryHelper.storePlatforms(telemetryProperties, "actuallyBuilt", platforms, settings);
+
+                return Settings.operateOnPlatforms(platforms,
```
digeff: If the implicit type works here, I'd remove the Q.Promise<any> return value to make these lines easier to read

src/taco-cli/cli/build.ts
MSLaguana: I was going to do this originally, but the line got very long and if I was going to split it over multiple lines I may as well try to increase readability.

src/taco-cli/cli/run.ts
MSLaguana: We are trying to move towards more explicit typings, so that's why I've added them here. Just spoke with @nisheetjain and he agrees that even though it isn't currently complained about by tslint, it's worth adding just to be sure.

src/taco-cli/cli/utils/settings.ts
```diff
@@ -102,6 +102,30 @@ class Settings {
...
+        localFunction: (platforms: string[]) => Q.Promise<any>,
+        remoteFunction: (platforms: string) => Q.Promise<any>): Q.Promise<any> {
+        var localPlatforms = platforms.filter((platform: Settings.IPlatformWithLocation) => {
```
digeff: Can you move BuildTelemetryHelper.extractPlatformsList to this file, and call that method?

src/taco-cli/cli/emulate.ts
```diff
@@ -128,16 +128,11 @@ class Emulate extends commands.TacoCommandBase {
...
-                        case Settings.BuildLocationType.Remote:
-                            // Just run remote, and failures are failures
-                            return Emulate.runRemotePlatform(platform.platform, commandData, telemetryProperties);
-                    }
-                }));
+
+                return Settings.operateOnPlatforms(platforms,
+                    (localPlatforms: string[]): Q.Promise<any> => CordovaWrapper.emulate(commandData, localPlatforms),
+                    (remotePlatform: string): Q.Promise<any> => Emulate.runRemotePlatform(remotePlatform, commandData, telemetryProperties)
+                    );
```
digeff: Shouldn't this parenthesis be on the previous line?

src/taco-cli/cli/utils/settings.ts
digeff: +1

MSLaguana: Any suggestions for where you think this function should go?

src/taco-cli/cli/build.ts
joshuaskelton: +1 in favor for readability.

src/taco-cli/cli/emulate.ts
MSLaguana: It can be; I thought it looked a little nicer and symmetric here.

src/taco-cli/cli/utils/settings.ts
dlebu: Create a new utility class if you don't have one already?

digeff: It seems to be highly related to the Platform concept, so why not PlatformUtilities? or PlatformOperations?

src/taco-cli/cli/build.ts
```diff
@@ -143,29 +144,22 @@ class Build extends commands.TacoCommandBase {
...
-                            return CordovaWrapper.build(commandData, platform.platform);
-                        case Settings.BuildLocationType.Remote:
-                            // Just build remote, and failures are failures
-                            return Build.buildRemotePlatform(platform.platform, commandData, telemetryProperties);
-                        default:
-                            return Q.reject(errorHelper.get(TacoErrorCodes.CommandBuildInvalidPlatformLocation, platform.platform));
-                    }
-                });
-            }));
+            var cleanPromise = Q({});
```
meenakb: var cleanPromise -> var cleanPromise: Q.Promise<any> in preparation for embracing tslint.

src/taco-cli/cli/utils/platformHelper.ts
```diff
@@ -0,0 +1,220 @@
...
+                return filteredPlatforms;
+            } else {
+                throw errorHelper.get(TacoErrorCodes.ErrorNoPlatformsFound);
+            }
+        });
+    }
+
+    public static determineSpecificPlatformsFromOptions(options: commands.ICommandData, settings: Settings.ISettings): PlatformHelper.IPlatformWithLocation[] {
+        var optionsToIgnore = options.original.indexOf("--") === -1 ? [] : options.original.slice(options.original.indexOf("--"));
+        var platforms = options.remain.filter(function (platform: string): boolean { return optionsToIgnore.indexOf(platform) === -1; });
```
meenakb: Use parseRequestedPlatforms() instead?

MSLaguana: Ah yes, thanks for catching this! A casualty of the refactoring.

meenakb:
:+1:

PR #51, opened 2015-10-02
Update cordova passthrough

joshuaskelton:


src/taco-cli/cli/utils/cordovaWrapper.ts
```diff
@@ -26,62 +26,64 @@ import projectHelper = require ("./projectHelper");
 import resources = require ("../../resources/resourceManager");
 import TacoErrorCodes = require ("../tacoErrorCodes");
 import errorHelper = require ("../tacoErrorHelper");
-import tacoUtility = require ("taco-utils");
+import tacoUtility = require("taco-utils");
```
MSLaguana: Does tslint complain about this?

src/taco-cli/cli/utils/cordovaWrapper.ts
```diff
@@ -188,6 +190,24 @@ class CordovaWrapper {
         options: { logLevel?: tacoUtility.InstallLogLevel, isSilent?: boolean, captureOutput?: boolean } = {}): Q.Promise<T | string> {
         return CordovaHelper.tryInvokeCordova<T | string>(apiFunction, () => CordovaWrapper.cli(processArgs(), options.captureOutput), options);
     }
+
+    /**
+     * Returns a path to the local Cordova command if ran from a taco project, otherwise it will return the
+     * global Cordova command.
+     */
+    private static getCordovaCommand(): Q.Promise < string > {
```
MSLaguana: Maybe rename to getCordovaPath? My first reaction to the name was that it would return a command intended for cordova, rather than returning which cordova should be spawned.

src/taco-cli/cli/utils/cordovaWrapper.ts
joshuaskelton: Strangely, no.

src/taco-cli/cli/utils/cordovaWrapper.ts
joshuaskelton: How about getCordovaExecutable?

MSLaguana: That sounds reasonable to me.

src/taco-cli/cli/utils/cordovaWrapper.ts
```diff
@@ -188,6 +190,24 @@ class CordovaWrapper {
...
+        return CordovaHelper.tryInvokeCordova<string>((cordova: typeof Cordova): string => {
+        return (<{ packageTargetPath?: string }> cordova).packageTargetPath;
```
nisheetjain: this is what we don't want. we should embed this in tacoPackageLoader. may be returned object has binPath?

MSLaguana: Are you referring to the typing specifically here? If so, we can fix that by making TacoPackageLoader functions return a type
T & {packageTargetPath?: string}
I believe (Or to be nicer, define a typedef LoadedPackage<T> with that type)

nisheetjain: no I am referring to Path.join here.. the logic to get to the executable should be part of TacoPackageLoader.

instead of exposing T.packageTargetPath, we should expose T.binPath or T.execPath.

Makes sense?

MSLaguana: Not every package has an executable though, while every package does have an install location. Do we want to tell TacoPackageLoader whether a package is executable (and what the name is), or should we leave that to the caller? I suppose we could read the package.json of the installed package and check for any "bin" entries.

nisheetjain: Yes, we should read the package.json. 
In fact, now when I think about it, this calls for a different API, say lazyRun("packageName@1.0.0". since we are not really requiring here

joshuaskelton: I like the idea of lazyRun, I'll look into heading in that direction.

src/taco-cli/cli/utils/cordovaHelper.ts
```diff
@@ -315,6 +318,23 @@ class CordovaHelper {
...
+    public static getCordovaExecutable(): Q.Promise<string> {
+        return projectHelper.getProjectInfo()
+            .then(function (projectInfo: projectHelper.IProjectInfo): any {
+                if (projectInfo.cordovaCliVersion) {
+                    return packageLoader.lazyRun(CordovaHelper.CORDOVA_NPM_PACKAGE_NAME, CordovaHelper.CORDOVA_NPM_PACKAGE_NAME + "@" + projectInfo.cordovaCliVersion, "cordova")
+                        .then(function (cordovaExecutablePath: string) {
+                            return cordovaExecutablePath;
+                        });
+                } else {
+                    return CordovaHelper.globalCordovaCommandName;
```
joshuaskelton: Not sure if I should reject here so we have a single place to handle returning the global cordova command.

src/taco-utils/tacoPackageLoader.ts
```diff
@@ -88,6 +89,7 @@ module TacoUtility {
 
     export interface ITacoPackageLoader {
         lazyRequire<T>(packageName: string, packageId: string, logLevel?: InstallLogLevel): Q.Promise<T>;
+        lazyRun(packageName: string, packageId: string, commandName: string): Q.Promise<string>;
```
meenakb: Would it be worthwhile to take an additional parameter for logLevel - instead of always defaulting to InstallLogLevel.warn?

nisheetjain: +1 to what @meenakb suggested. we can keep the API similar to lazyRequire...

joshuaskelton: Sure, I can make that change.

src/taco-utils/tacoErrorCodes.ts
```diff
@@ -18,6 +18,9 @@ module TacoUtility {
         PackageLoaderNpmUpdateFailedEaccess = 206,
         PackageLoaderNpmUpdateFailedWithCode = 207,
         PackageLoaderUpdateUnableToRecover = 208,
+        PackageLoaderRunPackageDoesntExportBinary = 209,
+        PackageLoaderRunPackageDoesntHaveRequestedBinary = 210,
+        PackageLoaderRunPackageBadBinaryFilePath = 211,
```
meenakb: Do these new error codes have associated strings in taco-utils\resources\en\resources.json? If so, they do not seem to be part of this PR :)

joshuaskelton: Yikes! Good catch.

src/taco-cli/cli/utils/cordovaHelper.ts
```diff
@@ -315,6 +318,23 @@ class CordovaHelper {
             });
     }
 
+    public static getCordovaExecutable(): Q.Promise<string> {
+        return projectHelper.getProjectInfo()
+            .then(function (projectInfo: projectHelper.IProjectInfo): any {
```
MSLaguana: This can return a `string | Q.Promise<string>` rather than `any`, to be a little more precise.

src/taco-cli/cli/utils/cordovaHelper.ts
```diff
@@ -315,6 +318,23 @@ class CordovaHelper {
...
+                if (projectInfo.cordovaCliVersion) {
+                    return packageLoader.lazyRun(CordovaHelper.CORDOVA_NPM_PACKAGE_NAME, CordovaHelper.CORDOVA_NPM_PACKAGE_NAME + "@" + projectInfo.cordovaCliVersion, "cordova")
+                        .then(function (cordovaExecutablePath: string) {
+                            return cordovaExecutablePath;
```
MSLaguana: This .then can be left off completely since it just returns its argument immediately

src/taco-utils/tacoPackageLoader.ts
```diff
@@ -96,6 +98,38 @@ module TacoUtility {
...
+
+                    if (!packageJson.bin) {
+                        return Q.reject<string>(errorHelper.get(TacoErrorCodes.PackageLoaderRunPackageDoesntExportBinary, packageName));
+                    }
+
+                    if (!packageJson.bin[commandName]) {
+                        return Q.reject<string>(errorHelper.get(TacoErrorCodes.PackageLoaderRunPackageDoesntHaveRequestedBinary, packageName, commandName));
+                    }
+
+                    var commandFilePath = path.join(request.targetPath, packageJson.bin[commandName]);
```
MSLaguana: Is this logic correct? My understanding was that if a package specified `bin: {"foo": "./bar.js"}` then it would create a node_modules/.bin/foo which was appropriately linked (foo.cmd on windows, symlinked with executable permissions on Linux/mac) to the ./bar.js file. With this logic, won't you be pointing to the .js file and try to run that, which will cause issues on windows?

joshuaskelton: Fixed the logic there.

MSLaguana:
Looks good to me.

src/typings/tacoPackageLoader.d.ts
```diff
@@ -27,6 +28,17 @@ declare module TacoUtility {
...
 
         /**
+         * Returns a path to the specified command exported from the specified package. If the package is not already downloaded,
+         * then first download and cache it locally.
+         *
+         * @param {string} packageName The name of the package to load
+         * @param {string} packageVersion The version of the package to load. Either a version number such that "npm install package@version" works, or a git url to clone
+         * @param {string} commandName The name of the binary to find
+         * @returns {Q.Promise<string>} A promise which is either rejected with a failure to find the local the binary or resolved with a path to the binary
+         */
```
guillaumejenkins: I think we usually put these headers with the implementation as well (in the *.ts file).

Otherwise I sign off on the PR!

joshuaskelton: Thanks! I'll be updating that in a moment.

PR #54, opened 2015-10-02
Fixed Android SDK install crash when updating the Tools package

guillaumejenkins:
We no longer update the Tools package as part of Android SDK configuration step. The Tools package already comes installed by default with the Android SDK zip file.

This essentially is a workaround for the current Android SDK crash when updating from Tools 24.3.3 to Tools 24.4 on Windows. This also speeds up the install-reqs process, and helps being future-proof against breaking changes, because we now use a static version of the Tools package.


nisheetjain:
:+1:  but this should go into Release branch instead...

PR #55, opened 2015-10-02
Now we show a note after reseting the certificates, informing the user that he needs to reset the running servers for the change to take effect

digeff:
Now the UX looks like:
> ./remotebuild certificates reset
Warning: This command deletes all current client certificates. You will need to generate and configure a new security PIN. Do you want to continue? [Y]es or [N]o
y
Note: You need to restart any running remote build servers for this change to take effect, and have the certificates be actually reseted
>


MSLaguana:
Looks good to me.

nisheetjain:
I agree with the rewording need here..

:+1:  otherwise

digeff:
Made the string change!

Issue #52, opened 2015-10-02
`taco create` fails to create project on Linux


omefire:
omefire@omefire-Linux:~$ taco create livereload-test

Downloading: /home/omefire/Projects/TACO/taco/build/packages/node_modules/taco-kits.tgz
npm WARN engine windows-no-runnable@0.0.6: wanted: {"node":"0.6"} (current: {"node":"0.10.25","npm":"2.13.5"})
taco-kits@1.0.0 node_modules/taco-kits
├── q@1.4.1
└── taco-utils@1.1.0 (rimraf@2.2.6, ncp@0.5.1, colors@1.1.2, applicationinsights@0.15.5, windows-no-runnable@0.0.6, unorm@1.3.3, semver@2.3.1, readline-sync@1.2.21, nopt@3.0.4, mkdirp@0.5.0, elementtree@0.1.6, lodash@3.10.1)

/home/omefire/Projects/TACO/taco/build/packages/node_modules/taco-kits.tgz downloaded and installed.
TACO2004: Error: The kit metadata file is not found. Please fix this by re-installing taco-kits from npm.

MSLaguana:
Looks like we've missed a filename casing issue!

PR #59, opened 2015-10-04
Fixing a typo in a filename

MSLaguana:


MSLaguana:
This is the fix for issue #52

nisheetjain:
Taco Kits tests still refer to the file with it's previous name. we need to fix those as well

MSLaguana:
Good point. Maybe we should add a Linux box to our test system as well to catch these issues.

nisheetjain:
:+1:

PR #60, opened 2015-10-05
Fixed Android SDK install crash when updating the Tools package

guillaumejenkins:
We no longer update the Tools package as part of Android SDK configuration step. The Tools package already comes installed by default with the Android SDK zip file.

This essentially is a workaround for the current Android SDK crash when updating from Tools 24.3.3 to Tools 24.4 on Windows. This also speeds up the install-reqs process, and helps being future-proof against breaking changes, because we now use a static version of the Tools package.

### DIFF
--- a/src/taco-dependency-installer/installers/androidSdkInstaller.ts
+++ b/src/taco-dependency-installer/installers/androidSdkInstaller.ts
@@ -21,6 +21,7 @@ import os = require ("os");
 import path = require ("path");
 import Q = require ("q");
 import request = require ("request");
+import util = require ("util");
 import wrench = require ("wrench");
 
 import InstallerBase = require ("./installerBase");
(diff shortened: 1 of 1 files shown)


src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -37,7 +38,7 @@ class AndroidSdkInstaller extends InstallerBase {
     private static AndroidHomeName: string = "ANDROID_HOME";
     private static AndroidCommand = os.platform() === "win32" ? "android.bat" : "android";
     private static AndroidPackages: string[] = [
-        "tools",
+        // "tools",  // Android SDK comes by default with the tools package, so there is no need to update it. In the future, if we feel we want dependency installer to always install the latest Tools package, then uncomment this.
```
MSLaguana: Maybe add a comment noting that this was causing issues? Is there a known issue in the android installer we can point to?

MSLaguana:
Looks good to me.

joshuaskelton:
Looks good.

meenakb:
:+1:

PR #63, opened 2015-10-06
Test fixes

MSLaguana:


src/taco-cli/package.json
```diff
@@ -54,7 +54,7 @@
         "gulp": "^3.8.10",
         "gulp-concat": "^2.4.2",
         "mocha": "latest",
-        "should": "*"
+        "should": "4.3.0"
```
nisheetjain: should we move to 7?

nisheetjain:
:+1:

src/taco-cli/test/create.ts
```diff
@@ -58,7 +58,7 @@ interface IKeyValuePair<T> {
 
 describe("taco create", function (): void {
     // Test constants
-    var createTimeout: number = 60000;
+    var createTimeout: number = 180000;
```
meenakb: Has this intentionally been increased to 180000?

src/taco-cli/package.json
meenakb: Why not "latest"? Is backward compatibility broken?

MSLaguana: I changed it to 4.3 since that is what all our other packages used at the moment (and it matched the typescript definitions we have). I had a look at the later versions, and we should be compatible with 6.X I believe, but 7.x has made breaking changes where instead of doing `foo.should.be.true` you need to do `foo.should.be.true()`. I didn't want to go change everything without discussing things first, but I did want to fix this.

src/taco-cli/test/create.ts
MSLaguana: Yes; I was still seeing occasional failures with 120000 seconds, so I bumped it up further. I think the main cause of timeouts is network instability at the moment.

joshuaskelton:
Looks good.

src/taco-cli/package.json
nisheetjain: we should create a bug/issue to track moving us to latest/7. This would help us leverage all the fixes and new features in should library

MSLaguana: I've created a bug to track the update to shouldjs 7.X

Issue #66, opened 2015-10-06
'taco run' silently fails on Linux


omefire:
Steps to repro:
- taco create proj
- cd proj
- taco run android

Behavior:
- does nothing

Expected:
- downloads and install android platform, then runs it.

Stepping through the code with node-inspector, I get the following error message: 
"Error: Cannot find module './utils/CordovaWrapper'"

MSLaguana:
Agh, another file casing issue.

MSLaguana:
taco-cli should work now. The tests are passing for me on an Ubuntu box.

PR #67, opened 2015-10-07
Fixing another filename capitalization issue.

MSLaguana:
Pull request #50 already solved issue #66, but while double checking all our require()s I found this one more that was incorrect.


omefire:
I think we have another issue around 'taco run'. it's around a module named 'remoteBuild' or something.

omefire:
Maybe, it would make sense trying taco-cli on a Unix-based OS, at least the main workflow: create, platform add, plugin add, run, etc...

That might go a long way in helping flush those capitalization issues.

MSLaguana:
I agree

MSLaguana:
Ok, with this commit I've actually gone and tested it out on a Linux box (Ubuntu specifically) and it was able to run tests, so it _should_ be good now.

omefire:
Awesome. Thanks Jimmy!

nisheetjain:
:+1:

meenakb:
:+1:

PR #74, opened 2015-10-09
Better aliasing for commands

joshuaskelton:
Working on a better way to resolve command/subcommand aliases.

### DIFF
--- a/src/taco-cli/cli/commands.json
+++ b/src/taco-cli/cli/commands.json
@@ -52,10 +52,7 @@
             "synopsis": "<synopsis></synopsis>",
             "categoryTitle": "[ProjectCommandsTitle]",
             "modulePath": "./templates",
-            "description": "<helptitle>[CommandTemplatesDescription]</helptitle>",
-            "aliases": {
-                "template": "templates"
-            }
+            "description": "<helptitle>[CommandTemplatesDescription]</helptitle>"
         }
         ,
         "kit":
@@ -203,7 +200,6 @@
             }
             ],
             "aliases": {
-                "platforms": "platform",
                 "rm": "remove",
                 "ls": "list"
             }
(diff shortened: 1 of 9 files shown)


src/taco-cli/cli/commands.json
```diff
@@ -265,19 +264,14 @@
...
-            ,
-            {
-                "alias": "ls", "command": "list"
+            "aliases": {
+                "plugins": "plugin",
+                "plugin": "plugin",
+                "remove": "remove",
+                "rm": "remove",
+                "list": "list",
+                "ls": "list"
```
MSLaguana: Should we include `"add": "add"` for completeness, even though it doesn't have any 'aliases' as such?

src/taco-cli/cli/commands.json
```diff
@@ -319,15 +313,12 @@
                 "description": "[CommandRemoteListSubcommandDescription]"
             }
             ],
-            "aliases": [
-            {
-                "alias": "rm", "command": "remove"
+            "aliases": {
+                "remove": "remove",
+                "rm": "rm",
```
MSLaguana: should be `"rm": "remove"`

joshuaskelton: Good catch.

src/taco-utils/helpCommandBase.ts
```diff
@@ -257,11 +264,10 @@ module TacoUtility {
          */
         private commandExists(command: string): boolean {
             for (var i in this.commandsFactory.listings) {
-                if (i === command) {
+                if (i === command || (this.commandsFactory.listings[i].aliases && (i === this.commandsFactory.listings[i].aliases[command]))) {
```
MSLaguana: This surprised me; I thought that command-level aliasing (template ~ templates, platform ~ platforms) was done based on the "aliases" entry of commands.json, but this does it by putting both command and subcommand aliases together in each command.

It works how it is at the moment, but I prefer the idea of having a single place where command-level aliases go.

MSLaguana:
Looks good to me.

PR #75, opened 2015-10-09
Improving checkForNewerVersion tests to handle errors correctly

MSLaguana:
checkForNewerVersion was hiding some test failures as timeouts.


nisheetjain:
:+1:

PR #77, opened 2015-10-10
Removing alias tests for now

joshuaskelton:


MSLaguana:
Do the original tests pass when you remove the aliasing tests?

joshuaskelton:
Again, the issue is how we test commands. We don't correctly construct them, and the aliasing is broken. I've created a task on myself to fix this.

joshuaskelton:
I've kicked off a test run for the changes. Looks good now.

MSLaguana:
If the issue was a change in the expected output, perhaps we can re-enabled the aliasing tests as well?

joshuaskelton:
I've updated the test expectations and added back the alias tests. Doing a run to verify the issue is fixed.

MSLaguana:
Looks good to me

PR #78, opened 2015-10-12
Code coverage

nisheetjain:
Sending this for early review. there is an existing issue where because we run coverage for all packages in same process, it tends to merge the coverages (same global.coverageVariable used)


nisheetjain:
BTW TACO_UNIT_TEST = false is needed to make sure tests don't pass the context around. this breaks remotebuild tests which compare against actual localized strings...

src/taco-cli/test/build.ts
```diff
@@ -103,6 +103,8 @@ describe("taco build", function (): void {
         kitHelper.kitPackagePromise = null;
         testHttpServer.close();
         rimraf(tacoHome, function (err: Error): void { done(); }); // ignore errors
+        // Reset mocked out resources
+        process.env["TACO_UNIT_TEST"] = false;
```
dlebu: Where is TACO_HOME reverted back? Do we care about it?

src/taco-cli/test/help.ts
```diff
@@ -59,14 +58,14 @@ describe("help for a command", function (): void {
...
         process.env["TACO_UNIT_TEST"] = true;
     });
 
     after(() => {
         // We just need to reset the stdout just once, after all the tests have finished
         process.stdout.write = stdoutWrite;
-        process.env["TACO_UNIT_TEST"] = previous;
+        // Reset mocked out resources
+        process.env["TACO_UNIT_TEST"] = false;
     });
```
dlebu: Does this change mean that we had tests not cleaning after running? Why do we force the "false" value and not the previous one anymore?

tools/GulpUtils.ts
```diff
@@ -19,51 +19,67 @@ import os = require ("os");
...
+        var coverageJsonFiles: string[] = [];
 
-                var modulePath = path.resolve(modulesRoot, val);
-                // check if package has any tests
-                var pkg = require(path.join(modulePath, "package.json"));
-                if (!pkg.scripts || !(GulpUtils.TestCommand in pkg.scripts)) {
-                    return Q({});
-                }
+        // for each module, sequentially run tests with coverage (with json reporting)  
+        return modulesToTest.reduce(function(soFar: Q.Promise<any>, val: string): Q.Promise<any> {
```
dlebu: Is there any way to make our tests run in parallel? That would make the run time a lot shorter.

src/taco-cli/test/build.ts
nisheetjain: what do you mean? this is TACO_UNIT_TEST

src/taco-cli/test/help.ts
nisheetjain: because now we are trying to run all the test cases in the same process. same as how we will run coverage..

tools/GulpUtils.ts
nisheetjain: not relevant for this PR...

nisheetjain:
codecoverage

this is how it looks

MSLaguana:
That's cool, though I wonder how it deals with things like the iOS build happening out of process; the bulk of the taco-remote-lib/ios lines of code will be in the iosBuild file which is run separately, so it might get undercounted. Similarly taco-dependency-installer...

src/taco-cli/test/build.ts
dlebu: What I meant was that TACO_HOME and TACO_UNIT_TEST used to be set and reset at the same time. Now that you modified the behavior for TACO_UNIT_TEST, do we have to do the same for TACO_HOME?

nisheetjain:
we are not trying to enforce to have x% of code coverage.. this is mostly for devs to get an idea what kind of coverage they have..

src/taco-cli/test/build.ts
nisheetjain: I guess. not part of this change. .test seems to be passing..

MSLaguana:
Right, I'm just wondering whether it is accurate for those child processes, or if we'll need to take those numbers with a larger grain of salt.

nisheetjain:
for now we need to take these numbers more as a guidance. for folks working on ios stuff should/can obtain these numbers on mac to get a more accurate picture..

this is a problem worth solving once we try to publicize these numbers on github or something

PR #80, opened 2015-10-12
Created first test for install-reqs using mockery (also updated some old tests, and create some testing infrastructure)

digeff:
- Created a single test for install-reqs
- Created the infrastructure to make creating that test, and other similar tests easy
    + Used mockery to mock the required modules
    + Created a module with some node fakes to use as mocks with mockery
- Changed the existing installerRunner tests to actually verify that no telemetry is expected, and use the new fake telemetry generator

### DIFF
--- a/src/taco-dependency-installer/test/androidSdkInstaller.ts
+++ b/src/taco-dependency-installer/test/androidSdkInstaller.ts
@@ -33,15 +33,15 @@ type TelemetryEvent = TacoUtility.ICommandTelemetryProperties;
 
 class FakeLogger implements ILogger {
     public log(message: string): void {
-        // TODO: Store all messages in memory, so we can test them
+        // Currently we don't care about these messages
     }
 
     public logWarning(message: string): void {
-        // TODO: Store all messages in memory, so we can test them
+        // Currently we don't care about these messages
     }
 
     public logError(message: string): void {
-        // TODO: Store all messages in memory, so we can test them
+        // Currently we don't care about these messages
     }
 
     public promptForEnvVariableOverwrite(message: string): Q.Promise<any> {
--- a/src/taco-dependency-installer/test/androidSdkInstaller.ts
+++ b/src/taco-dependency-installer/test/androidSdkInstaller.ts
@@ -1,0 +1,124 @@
+/**
+ *******************************************************
+ *                                                     *
+ *   Copyright (C) Microsoft. All rights reserved.     *
+ *                                                     *
+ *******************************************************
+ */
+
+/// <reference path="../../typings/dependencyInstallerInterfaces.d.ts"/>
+/// <reference path="../../typings/mocha.d.ts"/>
+/// <reference path="../../typings/mockery.d.ts"/>
+/// <reference path="../../typings/should.d.ts"/>
+
+"use strict";
+
+/* tslint:disable:no-var-requires */
+// var require needed for should module to work correctly
+// Note not import: We don't want to refer to shouldModule, but we need the require to occur since it modifies the prototype of Object.
+var shouldModule = require("should");
+/* tslint:enable:no-var-requires */
+
+import installerProtocol = require("../elevatedInstallerProtocol");
+import ILogger = installerProtocol.ILogger;
+import mockery = require("mockery");
+import nodeFakes = require("../../taco-tests-utils/nodeFakes");
+import Q = require("q");
+import _ = require("lodash");
+
+class FakeLogger implements ILogger {
+    public log(message: string): void {
+        // TODO: Store all messages in memory, so we can test them
+    }
+
+    public logWarning(message: string): void {
+        // TODO: Store all messages in memory, so we can test them
+    }
+
+    public logError(message: string): void {
+        // TODO: Store all messages in memory, so we can test them
+    }
+
+    public promptForEnvVariableOverwrite(message: string): Q.Promise<any> {
+        return Q({});
+    }
+}
+
+describe("AndroidSdkInstaller telemetry", () => {
+    var originalProcess = process;
+
+    before(() => {
+        // We tell mockery to replace "require()" with our own custom mock objects
+        mockery.enable({ warnOnUnregistered: false });
+    });
+
+    after(() => {
+        // Clean up and revert everything back to normal
+        process = originalProcess;
+        mockery.deregisterAll();
+        mockery.disable();
+    });
+
+    describe("updateVariablesDarwin", () => {
+        it("generates telemetry if there is an error on the update command", (done: MochaDone) => {
+            // We register our mocks. 1st process mocks, and then taco utils modules that require the process mocks
+            mockery.registerMock("child_process", new nodeFakes.FakeChildProcessModule().fakeAllExecCallsEndingWithErrors());
+
+            var tacoUtils = require("taco-utils"); // This needs to be required after mocking process
+            tacoUtils.Telemetry.init("TACO/dependencyInstaller", "1.2.3", false);
+
+            var telemetryTestsHelper = require("../../taco-tests-utils/telemetryTestsHelper");
+            var fakeTelemetryHelper = new telemetryTestsHelper.FakeTelemetryHelper();
+            var tacoUtilsWithFakeTelemetry = _.extend({}, tacoUtils, { TelemetryHelper: fakeTelemetryHelper, HasFakes: true });
+            mockery.registerMock("taco-utils", tacoUtilsWithFakeTelemetry);
+
+            // We require the AndroidSdkInstaller file, which will use all the mocked dependencies
+            var androidSdkInstallerClass = require("../installers/androidSdkInstaller");
+
+            var fakeProcess = new nodeFakes.FakeProcess().fakeMacOS().fakeDeterministicHrtime().asProcess();
+            // TODO: Try to find a nicer way of doing this. Maybe setting process as protoype of FakeProcess?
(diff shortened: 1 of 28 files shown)


src/taco-dependency-installer/test/androidSdkInstaller.ts
```diff
@@ -0,0 +1,124 @@
...
+/* tslint:disable:no-var-requires */
+// var require needed for should module to work correctly
+// Note not import: We don't want to refer to shouldModule, but we need the require to occur since it modifies the prototype of Object.
+var shouldModule = require("should");
+/* tslint:enable:no-var-requires */
+
+import installerProtocol = require("../elevatedInstallerProtocol");
+import ILogger = installerProtocol.ILogger;
+import mockery = require("mockery");
+import nodeFakes = require("../../taco-tests-utils/nodeFakes");
```
MSLaguana: You shouldn't have paths that escape the package. If you want to reference this file, it should be via a dev/optional dependency on the other package, I think.

src/taco-dependency-installer/test/androidSdkInstaller.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    describe("updateVariablesDarwin", () => {
+        it("generates telemetry if there is an error on the update command", (done: MochaDone) => {
+            // We register our mocks. 1st process mocks, and then taco utils modules that require the process mocks
+            mockery.registerMock("child_process", new nodeFakes.FakeChildProcessModule().fakeAllExecCallsEndingWithErrors());
+
+            var tacoUtils = require("taco-utils"); // This needs to be required after mocking process
+            tacoUtils.Telemetry.init("TACO/dependencyInstaller", "1.2.3", false);
+
+            var telemetryTestsHelper = require("../../taco-tests-utils/telemetryTestsHelper");
```
MSLaguana: Why require in here and not import up the top? Does this interact with the mocking?

src/taco-utils/promisesUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/// <reference path="../typings/lodash.d.ts" />
+
+"use strict";
+
+import _ = require("lodash");
+
+module TacoUtility {
+    export class PromisesUtils {
+        public static executeAfter<T>(valueOrPromise: T, afterCallback: { (): void }, failCallback?: { (reason: any): void }): T {
```
MSLaguana: I'm confused by what this function does. How is it different to Q(valueOrPromise).finally(afterCallback).fail(failCallback) ?

src/taco-dependency-installer/test/androidSdkInstaller.ts
digeff: yes, you need to call mockery.registerMock("child_process") before the call to require("child_process") is made for the mocking to work.

digeff: I'll try to organize this a little better in the next commit

src/taco-utils/promisesUtils.ts
digeff: The only difference is that a value won't be converted into a promise (this is used in telemetry.time) so:
var value = calculateSomethingSync();
can be converted to:
var value = telemetry.time(() => calculateSomethingSync()); // Here value is not a promise

And it also works for:
calculateSomethingAsync().then(doSomething);
which is converted to:
telemetry.time(() => calculateSomethingAsync()).then(doSomething); // Here the return value is a promise

src/taco-dependency-installer/test/androidSdkInstaller.ts
digeff: Converted to a new npm package!

src/taco-utils/promisesUtils.ts
MSLaguana: I don't think that's quite right, or at least not useful then:

If you are computing something synchronously, then
`executeAfter(someSyncCalculation(), () => console.log("done"), () => console.log("failed"))` is equivalent to `var result = someSyncCalculation(); executeAfter(result, () => console.log("done"), () => console.log("failed"))` and if the result isn't a promise, then it's just `var result = someSyncCalculation(); console.log("done")`. I don't see why you would invoke it like this...

digeff: Yes, for the synchronously case alone this method is useless.

The reason I use this method is because I wanted to have a single method that works both for sync and async cases instead of using telemetry.timeSync() and telemetry.timeAsync(), and this is the only way I could think of how to do it. If you come up with a different way of implementing this, I'm totally fine with changing it.

The other option would be to use 2 different methods with the implementations that you suggested:
telemetry.timeSync(): var result = someSyncCalculation(); afterCallback();
telemetry.timeAsync(): valueOrPromise.finally(afterCallback).fail(failCallback)

But I personally like having a single method for both cases. What do you think we should do?

digeff: Changed for a better implementation suggested by Jimmy

src/taco-dependency-installer/test/androidSdkInstaller.ts
```diff
@@ -0,0 +1,131 @@
...
+import tacoUtils = require("taco-utils");
+import _ = require("lodash");
+
+import nodeFakes = tacoTestsUtils.NodeFakes;
+
+type TelemetryEvent = TacoUtility.ICommandTelemetryProperties;
+
+class FakeLogger implements ILogger {
+    public log(message: string): void {
+        // TODO: Store all messages in memory, so we can test them
```
dlebu: If we really plan to test the messages, we should add an issue number here. If not, I would replace the "TODO" comments with "/* empty */"

src/taco-dependency-installer/test/androidSdkInstaller.ts
```diff
@@ -0,0 +1,131 @@
...
+                    step: { isPii: false, value: "updateVariables" },
+                    time: { isPii: false, value: "3000" },
+                    "updateVariables.time": { isPii: false, value: "2000" }
+                }
+            ];
+
+            androidSdkInstaller.run()
+                .done(() => done(new Error("Should have gotten a rejection in this test")), () => {
+                    fakeTelemetryHelper.getAllSentEvents().done((allSentEvents: TelemetryEvent[]) => {
+                        allSentEvents.should.eql(expectedTelemetry);
```
dlebu: Will the test time out if the assert fails?

src/taco-dependency-installer/test/androidSdkInstaller.ts
digeff: Changed to:
        // Currently we don't care about these messages

src/taco-dependency-installer/test/androidSdkInstaller.ts
digeff: No. I'm getting this when the assert fails:
  AndroidSdkInstaller telemetry
    updateVariablesDarwin

folder...\applicationinsights\AutoCollection\Exceptions.js:25
                    throw error;
                    ^
AssertionError: expected [
  {
  'initialStep.time': { isPii: false, value: '2000' },
  step: { isPii: false, value: 'initialStep' }
},
  {
  'error.description': {
  isPii: false,
  value: 'ErrorOnChildProcess INVALID MESSAGE on updateVariablesDarwin'
},
  lastStepExecuted: { isPii: false, value: 'updateVariables' },
  step: { isPii: false, value: 'updateVariables' },
  time: { isPii: false, value: '3000' },
  'updateVariables.time': { isPii: false, value: '2000' }
}
] to equal [
  {
  'initialStep.time': { isPii: false, value: '2000' },
  step: { isPii: false, value: 'initialStep' }
},
  {
  'error.description': {
  isPii: false,
  value: 'ErrorOnChildProcess on updateVariablesDarwin'
},
  lastStepExecuted: { isPii: false, value: 'updateVariables' },
  step: { isPii: false, value: 'updateVariables' },
  time: { isPii: false, value: '3000' },
  'updateVariables.time': { isPii: false, value: '2000' }
}
] (at '1' -> 'error.description' -> value, A has 'ErrorOnChildProcess INVALID MESSAGE on updateVariablesDarwin' and B has 'ErrorOnChildProcess on updateVariablesDarwin')
    at Assertion.fail (folder...\should\lib\assertion.js:113:17)
    at Assertion.prop.(anonymous function) [as eql] (folder...\should\lib\assertion.js:39:14)
    at folder...\test\androidSdkInstaller.js:99:42
From previous event:
    at folder...\test\androidSdkInstaller.js:98:56

dlebu: My concern was not the assert log, but the test status once the assert failed. Since your test is asynchronous, does the test actually fail immediately after the assert fails or ends up timing out, and fails because of that?

digeff: You were right, the test wasn't working.

Updating with this code which seems to work:
            return androidSdkInstaller.run()
                .then(() => Q.reject(new Error("Should have gotten a rejection in this test")), () => {
                    return fakeTelemetryHelper.getAllSentEvents().then((allSentEvents: TelemetryEvent[]) => {
                        allSentEvents.should.eql(expectedTelemetry);
                    });
                }).done(done, done);


Now I get:
 AndroidSdkInstaller telemetry
    updateVariablesDarwin
      1) generates telemetry if there is an error on the update command


  0 passing (63ms)
  1 failing

joshuaskelton:
Looks good to me.

PR #81, opened 2015-10-13
Changing version of taco-cli used in test

MSLaguana:
Changing the version from 5.2.0 to 4.3.0. This means that it matches the version used in other tests, and can be used from the cache rather than being redownloaded for this one test. Should improve test reliability.


digeff:
+1

PR #83, opened 2015-10-14
Adding a static class to properly construct command objects for tests.

joshuaskelton:
## Summary
The motivation for this change is to construct command objects for tests in the same way that we do for production code. In particular, alias testing needs command objects constructed with the commands.json file for correct aliasing.

## Changes
* Created static class CommandHelper which constructs the command object.
* Wherever we were new-ing command objects, a call to CommandHelper.getCommand is made.
* Clean up of no longer used imports.

### DIFF
--- a/src/taco-cli/test/run.ts
+++ b/src/taco-cli/test/run.ts
@@ -40,10 +40,10 @@ import BuildInfo = TacoUtility.BuildInfo;
 import Command = buildAndRunTelemetry.Command;
 import utils = TacoUtility.UtilHelper;
 
-import commandHelper = require ("./utils/commandHelper");
-import TacoCommandBase = TacoUtility.Commands.TacoCommandBase;
+import CommandHelper = require ("./utils/commandHelper");
+import ICommand = TacoUtility.Commands.ICommand;
 
-var create: TacoCommandBase = commandHelper.getCommand("create");
+var create: ICommand = CommandHelper.getCommand("create");
 
 describe("taco run", function (): void {
     this.timeout(60000); // The remote tests sometimes take some time to run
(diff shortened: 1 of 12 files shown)


MSLaguana:
Looks good to me.

src/taco-cli/test/utils/commandHelper.ts
```diff
@@ -0,0 +1,28 @@
...
+import commands = tacoUtility.Commands;
+import CommandsFactory = commands.CommandFactory;
+
+class CommandHelper {
+    private static commandsFactory: CommandsFactory = new CommandsFactory(path.join(__dirname, "../../cli/commands.json"));
+
+    /**
+     * Gets specified task object for use in testing.
+     */
+    public static getCommand(name: string): commands.TacoCommandBase {
```
nisheetjain: any reason why this shouldn't be ICommand?

src/taco-cli/test/run.ts
```diff
@@ -42,7 +40,10 @@ import BuildInfo = TacoUtility.BuildInfo;
 import Command = buildAndRunTelemetry.Command;
 import utils = TacoUtility.UtilHelper;
 
-var create: createMod = new createMod();
+import commandHelper = require ("./utils/commandHelper");
```
nisheetjain: please use PascalCase for static classes.

nisheetjain:
otherwise this looks good to me..

src/taco-cli/test/run.ts
joshuaskelton: Fixed.

src/taco-cli/test/utils/commandHelper.ts
joshuaskelton: Not that I can think of. This has been changed.

PR #84, opened 2015-10-16
Standalone android packages

guillaumejenkins:
We used to install Android Packages as part of Android SDK installation. This means that if Android SDK was already installed but didn't have the necessary packages, install-reqs would report a "Dependencies already installed" success, but the user still wouldn't be able to build his project.

Now, Android Packages are an independent "installer".


src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+        // Read the ANDROID_HOME from the process
+        var androidHomeValue: string = process.env[AndroidPackagesInstaller.ANDROID_HOME_NAME];
+
+        if (androidHomeValue) {
+            // ANDROID_HOME is set, use that to find the android executable
+            androidHomeValue = utilHelper.expandEnvironmentVariables(androidHomeValue);
+
+            // Build the android and adb commands and save them in members for reusability
+            var androidExecutable: string = os.platform() === "win32" ? "android.bat" : "android";
+            var adbExecutable: string = os.platform() === "win32" ? "adb.exe" : "adb";
```
dlebu: This pattern of constantly checking the platform in the code is making testing more difficult. @digeff had to mock the process object in his tests because of this. An easy way to fix this is to create an interface which contains all the platform specific details and to have two implementations, one for Windows and one for OS X. That way, if we add another platform later (say, Linux), we don't have to modify the code, but extend it (Open Closed Principle) See this for more info: http://www.objectmentor.com/resources/articles/ocp.pdf

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+                .add("error.description", "ErrorOnChildProcess on installAndroidPackages", /*isPii*/ false)
+                .addError(err);
+
+            if (err.code === "ENOENT") {
+                deferred.reject(resources.getString("AndroidCommandNotFound", path.basename(this.androidCommand)));
+            } else {
+                deferred.reject(err);
+            }
+        });
+        cp.on("exit", (code: number) => {
```
dlebu: Do we care about the exit code here if there is no error output?

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+        // Read the Android project properties
+        var projectPropertiesPath: string = path.join("platforms", "android", "project.properties");
+        var fileContent: string = fs.readFileSync(projectPropertiesPath).toString();
+
+        // Find the value at the "target=" line
+        var androidTargetId: string = /target=(.+)/.exec(fileContent)[1];
+
+        // Make sure the android target is in the list of available packages, otherwise it is an error state
+        if (this.availablePackages.indexOf(androidTargetId) === -1) {
+            throw new Error(resources.getString("UnknownAndroidTarget", androidTargetId, projectPropertiesPath));
```
dlebu: Should we introduce enum ids for error strings as well?

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+
+        // Call "android list sdk -e" and capture output
+        var args: string[] = [
+            "list",
+            "sdk",
+            "-e"
+        ];
+        var output: string = "";
+        var errorOutput: string = "";
+        var cp: childProcess.ChildProcess = os.platform() === "darwin" ?
```
dlebu: Same here.

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+        var deferred: Q.Deferred<any> = Q.defer<any>();
+        var args: string[] = [
+            "update",
+            "sdk",
+            "-u",
+            "--filter",
+            packagesFilter
+        ];
+        var errorOutput: string = "";
+        var cp: childProcess.ChildProcess = os.platform() === "darwin" ?
```
dlebu: Same here.

src/taco-dependency-installer/installers/androidSdkInstaller.ts
```diff
@@ -77,10 +67,6 @@ class AndroidSdkInstaller extends InstallerBase {
             });
     }
 
-    protected postInstallWin32(): Q.Promise<any> {
-        return this.postInstallDefault();
-    }
-
     protected downloadDarwin(): Q.Promise<any> {
```
dlebu: All these methods, downloadDarwin(), downloadWin32(), updateVariablesWin32(), updateVariablesDarwin(), etc, would go in the specific implementation instead of living in the same class.

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    constructor(installerInfo: DependencyInstallerInterfaces.IInstallerData, softwareVersion: string, installTo: string, logger: ILogger, steps: DependencyInstallerInterfaces.IStepsDeclaration) {
+        super(installerInfo, softwareVersion, installTo, logger, steps, "androidPackages");
+    }
+
+    protected installWin32(): Q.Promise<any> {
+        return this.installDefault();
+    }
+
+    protected installDarwin(): Q.Promise<any> {
```
dlebu: The install method is another good candidate for going into a new platform interface. Here, you call installDefault() for both, which is a clear sign that you are abstracting the implementation at a lower level, not at the higher level of your class architecture.

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
guillaumejenkins: These are not really TACOxxxx errors; we're just making some of the low-level installer errors clearer, so I would prefer to not have enum entries in TacoErrorCodes for them

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
guillaumejenkins: I agree. However, this is a big refactor that comes with a non-negligible cost (especially on the validation side). For that reason, I don't think it should be part of this changeset. I will create a backlog item for this so that it can be prioritized accordingly.

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
guillaumejenkins: We might as well; I will add this

src/taco-dependency-installer/installers/androidPackagesInstaller.ts
dlebu: That sounds good. Did we account for it already in the sprint planning today?

guillaumejenkins: I had created the task before the planning, but it didn't make it into this sprint

PR #85, opened 2015-10-19
Test Refactor and more "kit" command tests

meenakb:
Kit command tests


src/taco-cli/test/kit.ts
```diff
@@ -97,11 +108,30 @@ describe("Kit", function (): void {
...
+        .then(function (): Q.Promise<any> {
+                throw new Error("Scenario succeeded when it should have failed");
+        }, function (err: tacoUtils.TacoError): Q.Promise<any> {
+            err.errorCode.should.equal(expectedErrorCode);
+
+            // Also make sure that the project's taco.json
+            // file exists and has the expected values
+
+            fs.existsSync(tacoJsonPath).should.be.true;
+            var tacoJson: IKeyValuePair<string> = require(tacoJsonPath);
```
MSLaguana: Will tacoJsonPath ever be the same between two different tests in the same run? If so, I'd recommend against using `require` since that does caching based on file paths. To make sure you get the file as it exists on disk it may be better to explicitly `JSON.parse(fs.readFileSync(tacoJsonPath))`

joshuaskelton:
Looks good to me.

src/taco-cli/test/kit.ts
```diff
@@ -97,11 +108,30 @@ describe("Kit", function (): void {
 
             var tacoJson: IKeyValuePair<string> = require(tacoJsonPath);
 
-            tacoJsonKeyValues.should.be.eql(tacoJson);
+            tacoJsonKeyValues.should.be.eql(tacoJson)
```
digeff: Why did you remove the ; ?

src/taco-cli/test/kit.ts
```diff
@@ -198,8 +225,13 @@ describe("Kit", function (): void {
             rimraf(kitProjectpath, function (err: Error): void { done(); }); // ignore errors
         });
 
+        it("'taco kit select --cordova {Invalid-CLI-VERSION}' should execute with expected errors", function (done: MochaDone): void {
+            runKitCommandFailureCaseAndVerifyTacoJson<TacoErrorCodes>(["select", "--cordova", "InvalidCordovaCliVersion"], tacoJsonPath, expectedKitTacoJsonKeyValues, TacoErrorCodes.ErrorInvalidVersion)
+            .done(() => done(), done)
```
digeff: Isn't this equivalent to .done(done, done) ?

MSLaguana: No; `.done(done, done)` would pass through whatever arguments in the first case, so for example `[...].then(() => foo).done(done, done)` would invoke the first done with `foo` (which indicates an error condition). By doing `.done(() => done(), done)` instead we are explicitly invoking done with no arguments in the success case.

However last week I was reading the mocha documentation and noticed that we can avoid all of this by simply returning the promise, and mocha will succeed on a resolved promise and fail on a rejected promise.

digeff: Nice! I like the returning promise way!

src/taco-cli/test/kit.ts
meenakb: That's a good point. Although the project paths are different, fixed it.

src/taco-cli/cli/kit.ts
```diff
@@ -95,8 +104,9 @@ class Kit extends commands.TacoCommandBase {
         return Kit.promptUser(resources.getString("CommandKitSelectProjectUpdatePrompt"))
         .then(function (answer: string): Q.Promise<any> {
             if (answer && answer.length > 0 ) {
-                answer = answer.toLowerCase();
-                if (resources.getString("PromptResponseYes").split("\n").indexOf(answer) !== -1) {
+                // Try the lower-case version of the answer as well (This is needed so we can have better test coverage for "yes" and "no" response code paths)
```
MSLaguana: I don't follow this comment; why do we need to check both original-case and lower-case? why not just always convert to lower case as we did before?

src/taco-cli/test/kit.ts
```diff
@@ -64,6 +73,22 @@ describe("Kit", function (): void {
...
+
+    var expectedCliTacoJsonKeyValues2: IKeyValuePair<string> = {
+        "cordova-cli": "5.1.1"
+    };
+
+    var expectedKitTacoJsonKeyValues: IKeyValuePair<string> = {
+        kit: "5.1.1-Kit", "cordova-cli": "5.1.1"
+    };
+
+    var expectedKitPlatformVersion: IComponentVersionMap = {
```
MSLaguana: Why does this have a different type to the others? looks like it is actually equivalent, is that right?

src/taco-cli/cli/kit.ts
meenakb: In the test mode, we do not resolve the actual strings from resources.json (resouces.getString("PromptResponseYes") will just return "PromptResponseYes"  which means that in order to test the positive code paths, we will have to pass "PromptResponseYes" (for this check to succeed). The goal was to get as much test coverage as possible.

src/taco-cli/test/kit.ts
```diff
@@ -89,19 +114,96 @@ describe("Kit", function (): void {
...
+    function addPlatformToProject(platfromName: string, projectPath: string): Q.Promise<any> {
+        process.chdir(projectPath);
+        return platformRun(["add", platfromName]);
+    }
+
+    function addTestPluginsToProject(projectPath: string): Q.Promise<any> {
+        process.chdir(projectPath);
+        return pluginRun(["add", "cordova-plugin-camera"])
+        .then(function(): Q.Promise<any> {
+            return sleep(100);
```
MSLaguana: Why do we need a sleep here?

src/taco-cli/cli/kit.ts
MSLaguana: Ah, I see. In that case I'd be inclined to call toLowerCase on the getString() response as well, to make sure we are comparing both sides as lower case. I see that as less of a "making allowances for test scenarios" and more "protecting us against small mistakes in (possibly localized) strings", which also happens to mean we can work in a test scenario

src/taco-cli/test/kit.ts
```diff
@@ -89,19 +114,96 @@ describe("Kit", function (): void {
...
+                        break;
+                    default:
+                        errorHandler(new Error("Unexpected query!"));
+                }
+            },
+           close: onClose
+        };
+    }
+
+    function sleep(milliseconds: number): Q.Promise<any> {
```
MSLaguana: You can use Q.delay(milliseconds) for this

src/taco-cli/test/kit.ts
meenakb: Yes :) This can as well be of the same type.

src/taco-cli/test/kit.ts
meenakb: Q.delay(milliseconds) seems like a better alternative!

src/taco-cli/test/kit.ts
```diff
@@ -89,19 +114,96 @@ describe("Kit", function (): void {
...
-            var tacoJson: IKeyValuePair<string> = require(tacoJsonPath);
-
+            var tacoJson: IKeyValuePair<string> = JSON.parse(fs.readFileSync(tacoJsonPath, { encoding: "utf-8" }));
             tacoJsonKeyValues.should.be.eql(tacoJson);
             return telemetryParameters;
         });
     }
 
+    function runKitCommandFailureCaseAndVerifyTacoJson<T>(args: string[],
+        tacoJsonPath: string, tacoJsonKeyValues: IKeyValuePair<string>, expectedErrorCode: T): Q.Promise<TacoUtility.ICommandTelemetryProperties> {
```
MSLaguana: This function doesn't really return a Q.Promise<TacoUtility.ICommandTelemetryProperties> since it is only the error case. Perhaps Q.Promise<void> ? Then the Q.resolve(null) can be omitted and the final .then handlers can just return void

src/taco-cli/test/kit.ts
```diff
@@ -198,34 +302,45 @@ describe("Kit", function (): void {
...
-            runKitCommandAndVerifyTacoJsonContents(["select", "--cordova", "5.1.1"], tacoJsonPath, expectedCliTacoJsonKeyValues)
-                .then((telemetryParameters: TacoUtility.ICommandTelemetryProperties) => {
+        it("'taco kit select --cordova {Invalid-CLI-VERSION}' should execute with expected errors", function (done: MochaDone): void {
+            runKitCommandFailureCaseAndVerifyTacoJson<TacoErrorCodes>(["select", "--cordova", "InvalidCordovaCliVersion"], tacoJsonPath, expectedKitTacoJsonKeyValues, TacoErrorCodes.ErrorInvalidVersion)
+                .done(() => done(), done);
+        });
+
+        it("'taco kit select --cordova {CLI-VERSION}' on a project with a platform added, should execute with no errors", function (done: MochaDone): void {
+            /* tslint:disable:no-empty */
+            KitMod.yesOrNoHandler = getMockYesOrNoHandler(done, () => {}, "PromptResponseNo");
```
MSLaguana: We have a TacoUtils.UtilHelper.emptyMethod for intentionally empty functions so you don't need to disable the tslint check around them. Makes it clear that it is intended to be empty

src/taco-cli/test/kit.ts
```diff
@@ -198,34 +302,45 @@ describe("Kit", function (): void {
...
         var tacoJsonPath: string = path.resolve(cliProjectpath, "taco.json");
-        var expectedKitTacoJsonKeyValues: IKeyValuePair<string> = {
-            kit: "5.1.1-Kit", "cordova-cli": "5.1.1"
-        };
 
-        this.timeout(60000);
+        this.timeout(180000);
 
         before(function (done: MochaDone): void {
+            rimraf.sync(cliProjectpath);
```
MSLaguana: will cliProjectpath always exist? Or could this throw an exception if the folder is already absent?

src/taco-cli/test/kit.ts
```diff
@@ -233,16 +348,29 @@ describe("Kit", function (): void {
             rimraf(cliProjectpath, function (err: Error): void { done(); }); // ignore errors
         });
 
-        it("'taco kit select --kit {kit-ID}' should execute with no errors", function (done: MochaDone): void {
-            runKitCommandAndVerifyTacoJsonContents(["select", "--kit", "5.1.1-Kit"], tacoJsonPath, expectedKitTacoJsonKeyValues)
-                .then((telemetryParameters: TacoUtility.ICommandTelemetryProperties) => {
-                    var expected: TacoUtility.ICommandTelemetryProperties = {
+        it("'taco kit select --kit {Invalid-kit-ID}' should execute with expected error", function (done: MochaDone): void {
+            runKitCommandFailureCaseAndVerifyTacoJson<TacoKitsErrorCodes>(["select", "--kit", "InvalidKit"], tacoJsonPath, expectedCliTacoJsonKeyValues2, TacoKitsErrorCodes.TacoKitsExceptionInvalidKit)
```
MSLaguana: A better approach to tests which succeed or fail as a promise chain is to just return the promise, and not accept a done function. Just found this out recently, but it cleans things up a bunch.

This would then be
`it("...", function (): Q.promise<any> { return runKitCommandCaseAndVerifyTacoJson(...);})`

src/taco-cli/test/kit.ts
```diff
@@ -89,19 +110,92 @@ describe("Kit", function (): void {
...
+            return sleep(100);
+        }).then(function (): Q.Promise<any> {
+            var medicPluginTestsPath = path.resolve(".", "plugins", "cordova-plugin-camera", "tests");
+            var pluginCommandArgs: string[] = ["add", medicPluginTestsPath];
+            return pluginRun(pluginCommandArgs);
+        });
+    }
+
+    function getMockYesOrNoHandler(errorHandler: (err: Error) => void, onClose: () => void, desiredResponse: string): {
+        question: (question: string, callback: (answer: string) => void) => void;
```
MSLaguana: This could return an IMockReadLine if you export that from kit.ts

src/taco-tests-utils/projectHelper.ts
```diff
@@ -0,0 +1,93 @@
...
+                }
+
+                var cmdPath: string = path.join(projectPath, "platforms", platform, "cordova", cmdName);
+                var versionProc: child_process.ChildProcess = child_process.spawn(cmdPath);
+                versionProc.stdout.on("data", function(data: any): void {
+                    var version: string = data.toString();
+                    if (!path) {
+                        deferred.reject("Verson string not found");
+                    } else {
+                        version.trim().should.be.equal(platformsExpected[platform]);
```
MSLaguana: This check should probably not be an assertion, or it should be in a try catch, since right now I think this will be an uncaught exception

src/taco-tests-utils/projectHelper.ts
```diff
@@ -0,0 +1,93 @@
...
+                });
+                versionProc.on("error", function(err: any): void {
+                    deferred.reject(err);
+                });
+                return deferred.promise;
+            }));
+        }
+
+        public static checkPluginVersions(pluginsExpected: IKeyValuePair<string>, projectPath: string): void {
+            var deferred: Q.Deferred<any> = Q.defer<any>();
```
MSLaguana: This deferred isn't used

src/taco-cli/test/kit.ts
meenakb: Addressed the comments and added comments where appropriate.

src/taco-tests-utils/projectHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+module TacoTestsUtils {
+    interface IKeyValuePair<T> {
+        [key: string]: T;
+    }
+
+    export class ProjectHelper {
+        public static checkPlatformVersions(platformsExpected: IKeyValuePair<string>, projectPath: string): Q.Promise<any> {
+            var platformsInstalled: string[] = ProjectHelper.getInstalledPlatforms(platformsExpected, projectPath);
+            var onWindows: boolean = process.platform === "win32";
+            var deferred: Q.Deferred<any> = Q.defer<any>();
```
MSLaguana: I think this deferred should be inside the platformsInstalled.map function, otherwise as soon as the first platform finishes everything will be considered done (since there is only one deferred, returned multiple times)

MSLaguana:
Looks good other than that.

PR #87, opened 2015-10-19
Generate builds

nisheetjain:


src/taco-remote-lib/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "taco-remote-lib",
     "description": "Cordova-specific implementation of taco-remote's build functionality",
-    "version": "1.1.0",
+    "version": "1.1.0-dev",
```
MSLaguana: Shouldn't this be 1.1.1-dev now?

nisheetjain: thanks for pointing this out.. I didn't account for tslint change which affected all the packages.

tools/GulpUtils.ts
```diff
@@ -246,6 +198,151 @@ class GulpUtils {
...
+                        var entry: IDynamicDependencyEntry = json[packageKey];
+                        if (entry.dev) {
+                            switch (buildType) {
+                                case "dev":
+                                    entry.localPath = GulpUtils.getDevDependencyValue(entry.packageName, srcPath, destPath);
+                                    break;
+                                case "release":
+                                    entry.dev = undefined;
+                                    entry.localPath = undefined;
+                                    entry.packageId = entry.packageName;
```
MSLaguana: This isn't quite right I think. It'll work for now, but in future if we start having e.g. taco-remote-lib@1.2.3 and taco-remote-lib@2.0.0 supporting different sets of functionality, we don't want to strip out that side of things. Even if we have frozen 1.2.3 so only 2.0.0 has the dev flag, we want to point to 2.0.0 specifically I believe.

MSLaguana:
Looks good otherwise.

MSLaguana:
Oh, don't forget we need to update package versions in dynamicDependencies.json if they specify a specific version. I think this is just taco-remote-multiplexer's at the moment.

PR #88, opened 2015-10-19
Improving resiliency around iOS device launch

MSLaguana:
There was the potential for socket communication between us and the iOS device to have problems that we weren't catching, which could take down the whole server. We now listen to the error event of the communication, which should prevent that.
I also tried to fix up some error reporting problems that I encountered while trying to cause network communication issues.

Also fixing up how deleted files are handled, to make sure that platform-specific path separators are properly handled and we also don't allow files outside of the project directory to be deleted.


src/taco-remote/lib/buildManager.ts
```diff
@@ -369,10 +369,20 @@ class BuildManager {
         if (fs.existsSync(changeListFile)) {
             buildInfo.changeList = JSON.parse(fs.readFileSync(changeListFile, { encoding: "utf-8" }));
             if (buildInfo.changeList) {
+                buildInfo.changeList.deletedFiles = buildInfo.changeList.deletedFiles.map(function (deletedFile: string): string {
+                    // Convert all \ and / characters in the path string into platform-appropriate path separators
+                    return deletedFile.split("\\").join(path.sep).split("/").join(path.sep);
```
nisheetjain: wouldn't the split break if this code runs on Windows instead?

src/taco-remote/lib/buildManager.ts
```diff
@@ -369,10 +369,20 @@ class BuildManager {
...
+                });
                 buildInfo.changeList.deletedFiles.forEach(function (deletedFile: string): void {
-                    if (!deletedFile.match(/^plugins[\/]/)) {
+                    
+                    if (deletedFile.split(path.sep)[0] === "plugins") {
```
nisheetjain: shouldn't the comparison be !== if we want to skip plugins scenario? how do we catch scenario where request is foo\..\plugins\?

src/taco-remote/lib/buildManager.ts
MSLaguana: I don't think that the split would break things, but I've changed to a replace since it's a bit nicer anyway.

src/taco-remote/lib/buildManager.ts
MSLaguana: Clients should not be sending sending paths like foo/../bar/../baz, so if they do so we will try not to do anything bad, but I don't think it is necessary to be perfect there. If we do resolve, we still might end up with something like "../cordovaApp/plugins" which (in the context of where it is eventually used) is equivalent to "plugins" but that's harder to catch, so we'd need to do something like resolve it to an absolute location, then re-relativize it, and then see if it started with plugins.

dlebu:
LGTM

dlebu:
I closed and reopened this by mistake. Change LGTM.

PR #91, opened 2015-10-20
Added full coverage tests for the telemetry code in AndroidSdkInstaller

digeff:
- Added full coverage tests for the telemetry code in AndroidSdkInstaller
- Added mock-fs to mock the file system
- Added mock-spawn to mock child_process.spawn
- Replaced process with tacoUtils.ProcessUtils.getProcess() in more places, so we can easily mock it
- Added some improvements to the fake objects

Tests
All test pass. Tested install-reqs on an empty Windows 10 machine and it worked, I was able to build on Android after that.


src/taco-dependency-installer/test/androidSdkInstaller.ts
```diff
@@ -121,12 +158,193 @@ describe("AndroidSdkInstaller telemetry", () => {
...
+                },
+                {
+                    "error.description": { isPii: false, value: "NeedInstallDestination on installDefault" },
+                    "install.time": { isPii: false, value: "2000" },
+                    lastStepExecuted: { isPii: false, value: "install" },
+                    step: { isPii: false, value: "install" },
+                    time: { isPii: false, value: "3000" }
+                }];
+
+            return telemetryGeneratedShouldBe(expectedTelemetry, /An installation destination is needed for this dependency/, done);
```
guillaumejenkins: We should set the resource manager to test mode in the before() handler so that we only have to deal with string IDs, not full strings

digeff: That's a good idea! Fixed!

PR #92, opened 2015-10-20
Updating to ios-sim version 5

MSLaguana:
ios-sim 5 uses a different approach to specify a simulation target, but we still support all the same previous inputs. As part of this change we also now allow the iOS version to be specified, but if the target/version combination doesn't exist on the server then it will fail.

### DIFF
--- a/src/taco-remote-lib/ios/iosEmulateHelper.ts
+++ b/src/taco-remote-lib/ios/iosEmulateHelper.ts
@@ -23,7 +23,7 @@ import utils = tacoUtils.UtilHelper;
 import BuildInfo = tacoUtils.BuildInfo;
 
 // Note: this file is not intended to be loaded as a module, but rather in a separate process.
-process.on("message", function (emulateRequest: { appDir: string; appName: string; target: string }): void {
+process.on("message", function (emulateRequest: { appDir: string; appName: string; target: string, version: string }): void {
     Q(IOSEmulateHelper.emulate(emulateRequest))
         .then(function (result: { status: string; messageId: string; messageArgs?: any }): void {
         process.send(result);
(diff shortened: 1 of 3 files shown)


nisheetjain:
don't we need a new taco-remote-lib (new entry in mux) for this?

MSLaguana:
This fix is backwards compatible (and actually addresses bugs in previous versions) so we don't need a separate entry in taco-remote-multiplexer, just need to update it when we get to publishing. I'm waiting to update the package versions until we take the RI from v1.0.2 so we don't have too many conflicts.

src/taco-remote-lib/ios/iosEmulateHelper.ts
```diff
@@ -32,17 +32,19 @@ process.on("message", function (emulateRequest: { appDir: string; appName: strin
...
     private static IOS_SIMULATOR_TARGETS: { [id: string]: string } = {
-        "iphone 4s": "--retina",
-        "iphone 5": "--retina --tall",
-        "iphone 5s": "--retina --tall --64bit",
-        "iphone 6": "--devicetypeid com.apple.CoreSimulator.SimDeviceType.iPhone-6",
-        "iphone 6 plus": "--devicetypeid com.apple.CoreSimulator.SimDeviceType.iPhone-6-Plus",
-        "ipad 2": "--family ipad",
-        "ipad air": "--family ipad --retina --64bit",
-        "ipad retina": "--family ipad --retina"
+        "iphone 4s": "iPhone-4s",
```
nisheetjain: so we don't need these flags (--retina) anymore?

MSLaguana: Yeah, ios-sim 4 (and 5) changed how simulator targets are specified. You now specify `--devicetypeid 'simulatorTarget, iosVersion'` e.g. `--devicetypeid 'iPad-2, 9.0'`

nisheetjain:
:+1:

PR #94, opened 2015-10-20
Bugjail guillaume

guillaumejenkins:
-Removed duplicate message when creating a project
-TACO now deletes files to ignore from the temporary template folder before creating the project, so Cordova no longer copies files we want to ignore


nisheetjain:
:+1:

PR #97, opened 2015-10-21
Bumping version numbers for next dev cycle

MSLaguana:
Also fixing a tslint issue and a bug in how dynamicDependencies.json was modified in dev


nisheetjain:
:+1:

PR #98, opened 2015-10-21
Fixing an error in tacoPackageLoader where we cached a value we shouldn't

MSLaguana:


nisheetjain:
:+1:

src/taco-utils/tacoPackageLoader.ts
```diff
@@ -322,7 +322,7 @@ module TacoUtility {
         private static createTacoPackageInstallRequest(packageKey: string, dependencyConfigPath: string, logLevel: InstallLogLevel): IPackageInstallRequest {
             if (fs.existsSync(dependencyConfigPath)) {
                 try {
-                    var dependencyLookup: any = require(dependencyConfigPath);
+                    var dependencyLookup: any = JSON.parse(<any> fs.readFileSync(dependencyConfigPath));
```
digeff: I think it'd be safer to pass an encoding, or assume that you get back a buffer. According to this: https://github.com/nodejs/node-v0.x-archive/issues/186 that call to readFileSync will return a buffer, and that matches our .d.ts file:
    export function readFileSync(filename: string, encoding: string): string;
    export function readFileSync(filename: string, options: { encoding: string; flag?: string; }): string;
    export function readFileSync(filename: string, options?: { flag?: string; }): Buffer;

MSLaguana: It does return a buffer, and JSON.parse can handle a buffer, but our typings for JSON.parse don't acknowledge that so it would complain. The code runs fine though.

digeff: Can you please put your response as a comment, so we know that we can remove the <any> when we get a newer .d.ts file with the proper typings?

After that :+1:

MSLaguana: We have a lot of JSON.parse's throughout the code with similar workarounds already

digeff: Maybe we should create an issue instead.

Issue #96, opened 2015-10-21
RemoteBuild Hangs with Node 4.0


nisheetjain:
Issue was reported by a customer here
http://stackoverflow.com/questions/33225961/ios-remote-build-infinite-attempt-on-visual-studio-cordova/33247768#33247768

Taco should proactively detect the scenario and fail the build with clear error messages

MSLaguana:
We've got the error message in now and we won't deadlock remotebuild anymore, so I'm closing this issue.

PR #99, opened 2015-10-21
Adding a check for incompatible node/cordova versions

MSLaguana:
Cordova <5.3.3 breaks on node >=4.0.0 if the deployment-target or target-device preferences are specified in config.xml.
When run under remotebuild, this shows up as a deadlock which holds up all further processing. Now we check for
this combination and proactively reject the build with a message to the end user.
Addresses issue #96

### DIFF
--- a/src/taco-remote-lib/resources/en/resources.json
+++ b/src/taco-remote-lib/resources/en/resources.json
@@ -68,7 +68,7 @@
     "_UnsupportedPlatform.comment": "Message reported to the client when a request for an unsupported platform is made",
     "NoCordovaVersionSpecified": "No Cordova version specified!",
     "_NoCordovaVersionSpecified.comment": "Error reported when trying to build without any version specified",
-    "UnsupportedCordovaAndNodeVersion": "Unable to build using Cordova earlier than 5.3.3 when on node version 4.0.0 and higher and also specifying deployment-target or target-device preferences in config.xml",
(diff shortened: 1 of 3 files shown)


src/taco-remote-lib/resources/en/resources.json
```diff
@@ -68,6 +68,8 @@
     "_UnsupportedPlatform.comment": "Message reported to the client when a request for an unsupported platform is made",
     "NoCordovaVersionSpecified": "No Cordova version specified!",
     "_NoCordovaVersionSpecified.comment": "Error reported when trying to build without any version specified",
+    "UnsupportedCordovaAndNodeVersion": "Unable to build using Cordova earlier than 5.3.3 when on node version 4.0.0 and higher and also specifying deployment-target or target-device preferences in config.xml",
```
nisheetjain: strings need to be reviewed.
May be "Unable to build incompatible configuration: 4.0<= cordova < 5.3, Node < 4.0, deployment-target: true, target-device"

MSLaguana: @subhagpo can you have a look at this string?

subhagpo:
How about - 

Incompatible configuration: Unable to build using Cordova <= 5.3.3 and Node >= 4.0.0, with deployment-target and target-device specified in the config.xml.

PR #101, opened 2015-10-21
Fixing file:// handling in tacoPackageLoader

MSLaguana:
Because variables were defined in the wrong order, the file regex was searching for "undefined" rather than "file://"

### DIFF
--- a/src/taco-utils/tacoPackageLoader.ts
+++ b/src/taco-utils/tacoPackageLoader.ts
@@ -93,12 +93,12 @@ module TacoUtility {
     }
 
     export class TacoPackageLoader {
+        public static FILE_REGEX_PREFIX: string = "file://";
+
         public static GIT_URI_REGEX: RegExp = /^http(s?)\\:\/\/.*|.*\.git$/;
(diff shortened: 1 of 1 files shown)


src/taco-utils/tacoPackageLoader.ts
```diff
@@ -93,13 +93,13 @@ module TacoUtility {
     }
 
     export class TacoPackageLoader {
+        public static FILE_REGEX_PREFIX: string = "file://";
```
nisheetjain: instead of making FILE_REGEX_PREFIX public, let's create a "test" API instead?

MSLaguana: The reason I made it public is it needs to be defined before FILE_URI_REGEX (which is public so other classes can use it), and if it is private then tslint complains that I have a private variable declared before a public one.

digeff: You can add a tslint exception

nisheetjain: @MSLaguana I know why this needs to be changed but instead of making it public, wouldn't it be better to roll this into a public API or may be add a tslint exception

nisheetjain:
:+1:

PR #103, opened 2015-10-21
Removing a platform should remove the corresponding entry from config.xml

meenakb:


nisheetjain:
:+1:

MSLaguana:
I see a check for whether the user has done "taco platform add platform@version", but that check will also apply to "taco platform rm platform@version". Cordova seems to basically do nothing if you specify such an argument, do we behave appropriately?

meenakb:
"cordova platform remove platform@version" is a no-op. TACO should and does not update the platforms or config.xml in this case.

MSLaguana:
Sounds good. Looks good to me.

PR #104, opened 2015-10-22
Adds support for creating an external telemetry object.

TimBarham:
Defines an object that can be passed to external modules (like `taco-simulate`) to report back telemetry events. It has a single method - `sendTelemetry()`.

Also, `sendCommandSuccessTelemetry()` and `sendCommandFailureTelemetry()` `args` parameter should be declared as optional, since it has a default value specified.

### DIFF
--- a/src/taco-utils/telemetryHelper.ts
+++ b/src/taco-utils/telemetryHelper.ts
@@ -28,11 +28,15 @@ module TacoUtility {
     export interface ITelemetryPropertyInfo {
         value: any;
         isPii: boolean;
-    };
+    }
 
     export interface ICommandTelemetryProperties {
         [propertyName: string]: ITelemetryPropertyInfo;
-    };
+    }
+
+    export interface IExternalTelemetryProvider {
+        sendTelemetry: (event: string, props: Telemetry.ITelemetryProperties, error?: Error) => void;
+    }
 
     interface IDictionary<T> {
         [key: string]: T;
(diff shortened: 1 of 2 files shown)


src/taco-utils/telemetryHelper.ts
```diff
@@ -246,6 +246,38 @@ module TacoUtility {
...
         }
 
+        /**
+         * Creates an object that can be passed to external modules to report telemetry events.
+         * @param componentName - a name to prefix on the telemetry event names received from the external module.
+         * @param baseProps - a collection of properties to include with all events.
+         * @param errorHandler - an optional function to wrap or modify any errors (wrap in a TacoError, for example).
+         * @returns {{sendTelemetry: (function(string, ITelemetryProperties, Error=): void)}}
+         */
+        public static getExternalTelemetryObject(componentName: string, baseProps: Telemetry.ITelemetryProperties, errorHandler?: (error: Error) => Error) : Object {
```
MSLaguana: Maybe we should define an interface for this return value, so it isn't just an opaque Object

TimBarham: The reason I didn't is that this is just being passed to some external (non-TACO) module, so there didn't seem to be any benefit. I actually originally had a comment to that effect, but must have deleted it when moving code around. Can you think of a scenario where having an interface defined would be useful? I guess if that external module was written in TypeScript it would be useful (currently it is only being used by JavaScript code).

MSLaguana: I see interfaces as being especially useful when they are at the boundary between modules, as a form of documentation. You've basically defined the type in the JSDoc above, may as well make it formal and let the type system know about it too.

digeff:
Have you considered passing around the SESSION_ID instead, and setting it? That way you could easily reuse all the utilities that we built on top of telemetry, like TelemetryGenerator

TimBarham:
@digeff - yeah, I went with this approach because ultimately this object is intended to be used by a module that is likely to be used by others who don't want to use our telemetry, but may be interested in supporting their own. So its design was driven by the needs of that module - completely generic and not dependent on TACO in any way - rather than what TACO could potentially support.

I originally implemented it within `simulate.ts` (where the `taco simulate` command is implemented), but figured it was actually generic enough to live here since there was the potential for it to be useful in other scenarios.

digeff:
I think it's great to let them generate their own telemetry, but I think we should still try to re-use as much as the infrastructure as we can so we don't go about reinventing the wheel.

The stuff we have so far supports any property you want to add to telemetry, so it's pretty generic. If for any reason that doesn't work for someone, we could talk with them and update the infrastructure to consider their use case, or they can update them themselves too.

TimBarham:
It's not that we want to let anyone generate their own telemetry - it's that other companies can use the plugin simulation module within their own products, and the simulation module supports providing telemetry data using their existing telemetry infrastructure.

So this is just TACO's implementation of a generic interface that the simulation module expects if it should provide telemetry data, and TACO's implementation uses TACO's existing telemetry infrastructure. Some other company that uses the simulation module in a product that already has its own telemetry infrastructure can provide an equivalent object, with the same interface, but a different implementation (that uses their infrastructure).

The simulation module itself is agnostic to the telemetry infrastructure of the product that is using it, and it only provides telemetry data if it is used within a product that requests it (by providing an object with this interface). It has no telemetry infrastructure itself - if this object is provided, it calls the `sendTelemetry()` method when it wants to report an event, and that's it.

src/taco-utils/telemetryHelper.ts
TimBarham: Ok, done.

MSLaguana:
Looks good to me

digeff:
Ah! I completely misunderstood that point, now I understand what you are doing/why, and it makes a lot of sense! :)

Looks good to me too (sorry, for the confusion)

TimBarham:
Thanks guys. And no worries @digeff - I figured our wires were getting crossed :smile:.

Issue #106, opened 2015-10-23
remotebuild is not extracting all files from the upload_BUILDNUM.tgz


sdhintGIT:
I had originally thought that my plugin folders/files were not getting copied from the windows machine over to the mac, but I found them inside the upload_BUILD.tgz they are just not getting extracted before building.

is there a configuration or some file i can update to make sure all files are extracted?

I installed remotebuild today on the mac, version 2.0.1 , running same node version as my windows machine, 0.12.2 

any help is appreciated, it's literally driving me nuts. vs-mda-remote was previously installed on this machine as i was an early adopter of cordova tools in vs2013.

MSLaguana:
We treat the /plugins folder as special when uploading files to make sure that we install/uninstall/update plugins appropriately, but there could be a bug here.

Instead of extracting directly to /plugins, we extract to /remote/plugins, and then go and invoke "cordova plugin add /remote/plugins/thePlugin" if it is a new plugin (if the plugin.xml is new) or we copy the files to /plugins/thePlugin if the plugin already existed.

Could you please give some more details about what's happening? Is nothing at all being added to /plugins/ ?

sdhintGIT:
On the mac in builds/BUILDNUM/cordovaApp  it creates a plugins folder and there is one plugin there.  cordova-plugin-whitelist, fetch.json, ios.json and .DS_Store
In /builds/BUILDNUM/cordovaApp/www there is no plugins folder

On the windows machine, it’s correctly built in platforms,
[cid:image001.png@01D10DB0.6622FFF0]

[cid:image002.png@01D10DB0.6622FFF0]


and is also in the .tgz on the mac, just not extracted. To cordovaApp


I can send you the .tgz if it would help with getting this fixed. I have to have this working.
I have tried:
Removing the ios platform through command line and reading it, removing all plugins and reinstalling them after blowing away the ios platform, still didn’t take.


It was working fine with vs2013 and vs-mda-remote for months and months.


_________________________

SAMANTHA ALDRIDGE
Software Architect
Finished Vehicle Logistics

502.570.3302  direct
859.433.3287  cell
502.570.4302 fax
Email saldridge@vascorlogistics.com<mailto:saldridge@vascorlogistics.com>

VASCOR, Ltd.
100 Farmers Bank Drive
Suite 300
Georgetown, KY 40324
www.vascorlogistics.com<http://www.vascorlogistics.com/>

sdhintGIT:
[...]

PR #108, opened 2015-10-24
Hotfix: Fixing a bug in extracting plugins

MSLaguana:
Addressing #106 on top of the previous release directly so we can get a fixed build out.

### DIFF
--- a/src/taco-remote/lib/buildManager.ts
+++ b/src/taco-remote/lib/buildManager.ts
@@ -285,6 +285,7 @@ class BuildManager {
     private beginBuild(req: express.Request, buildInfo: BuildInfo): void {
         var self: BuildManager = this;
         var extractToDir: string = path.join(buildInfo.buildDir, "cordovaApp");
+        var extractToRemoteDir = path.join(extractToDir, "remote");
         buildInfo.buildSuccessful = false;
         buildInfo.appDir = extractToDir;
         try {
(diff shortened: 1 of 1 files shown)


sdhintGIT:
how do I get the updated version of remotebuild? getting from npm still has this bug.

MSLaguana:
This pull request hasn't been reviewed yet; once we do that (soon hopefully) we'll publish an update.

src/taco-remote/lib/buildManager.ts
```diff
@@ -328,7 +329,7 @@ class BuildManager {
 
             // Here we want to exclusively extract the contents of the /plugins folder, and we will put it in a separate location
             // Later in taco-remote-lib we will manually merge the plugins into the project to ensure they are added correctly.
-            var localPath: string = path.relative(extractToDir, who.props.path);
+            var localPath: string = path.relative(extractToRemoteDir, who.props.path);
             return !who.props.depth || (who.props.depth === 0 && who.props.Directory) || localPath.split(path.sep)[0] === "plugins";
```
meenakb: Does taco-remote-lib look for the plugins in the right directory and behave as expected?

MSLaguana: Yes, all that side of things works correctly, it was just the initial extraction of files that was broken after a refactoring.

meenakb: :+1:  Looks good to me then.

nisheetjain:
:+1:

PR #109, opened 2015-10-24
Improving error handling with remotebuild

MSLaguana:
Some errors when attempting HTTP[S] communication will not have a "code" property, for example if a HTTPS connection is rejected because the server's hostname does not match the names on the certificate. This change make sure that we give appropriate errors rather than crashing.


digeff:
:+1:

joshuaskelton:
:+1:

PR #110, opened 2015-10-24
Fixing an issue with remotebuild and ipv6

MSLaguana:
When remotebuild responds to a new build submssion, it includes a URL where more information can be found. When the client is connecting to the server via an ipv6 address this URL becomes malformed (e.g. the ipv6 address [abcd:1234::5678]:8080 is abbreviated to abcd:8080) and there is no good way to address that on the server. Instead we have the client ignore the host part of the URL and substitute the host that it wants to communicate with.


digeff:
:+1: 
Does this crashes if response.headers["content-location"] is null? or an invalid url?

MSLaguana:
If there is an invalid URL then it should work fine (until we try to connect to it and that fails). If the content-location header isn't set then we'll throw an exception at the moment, but that is part of the contract with remotebuild.

nisheetjain:
:+1:

PR #112, opened 2015-10-26
Created telemetry tests for dependencyInstaller, installerRunner and javaJDKInstaller

digeff:
- Created new tests
- Refactored FakeLogger into it's own file
- Updated some node fakes


src/taco-dependency-installer/test/javaJdkInstaller.ts
```diff
@@ -0,0 +1,344 @@
...
+                realCallback(error, /*stdout*/ new Buffer(""), /*stderr*/ new Buffer(""));
+            }
+
+            callTime++;
+            return new nodeFakes.ChildProcess();
+        }
+
+        it("generates telemetry error when the package installer gives an error", (done: MochaDone) => {
+            // child_process.exec call will only fail for the install packages call
+            childProcessModule.exec = fakeExecToTestPackageInstallerFailure;
```
guillaumejenkins: Small nit but could `childProcessModule.exec = fakeExecToTestPackageInstallerFailure;` for these 3 tests be moved to the `beforeEach()`?

src/taco-dependency-installer/test/javaJdkInstaller.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    describe("installation in mac", () => {
+        enum ChildProcessExecBehavior {
+            PackageInstallerCommandFailsScenario,
+            PackageInstallerCommandCantBeRunScenario,
+            DetachingDmgFileFailsScenario
+        }
+
+        var callTime: number;
+        var childProcessExecBehavior: ChildProcessExecBehavior;
```
guillaumejenkins: Does it make sense to reset this to null in the `beforeEach()` or in a `afterEach()`? That way, if someone new adds a test case in the future and forgets to set it to the correct value in his test, we won't get undeterministic results?

src/taco-dependency-installer/test/javaJdkInstaller.ts
```diff
@@ -0,0 +1,344 @@
...
+                var couldntRunPackageInstallerError = new Error("Package installer exited with an error");
+                switch (childProcessExecBehavior) {
+                    case ChildProcessExecBehavior.PackageInstallerCommandFailsScenario:
+                        (<any> couldntRunPackageInstallerError).code = 54;
+                        break;
+                    case ChildProcessExecBehavior.PackageInstallerCommandCantBeRunScenario:
+                        break;
+                    case ChildProcessExecBehavior.DetachingDmgFileFailsScenario:
+                        couldntRunPackageInstallerError = null;
+                        break;
```
guillaumejenkins: Just for future-proofing the authoring of more test cases, I would add a default case in this switch with some generic error like "Bad test configuration". This goes with my above suggestion of reseting `childProcessExecBehavior` to null before/after each test.

src/taco-dependency-installer/test/javaJdkInstaller.ts
```diff
@@ -0,0 +1,344 @@
...
+                        value: "InstallDestination needed on installWin32"
+                    },
+                    "install.time": { isPii: false, value: "2000" },
+                    lastStepExecuted: { isPii: false, value: "install" },
+                    step: { isPii: false, value: "install" },
+                    time: { isPii: false, value: "3000" }
+                }
+            ];
+
+            return telemetryGeneratedShouldBe(expectedTelemetry, /An installation destination is needed for this dependency/, done);
```
guillaumejenkins: `/An installation destination is needed for this dependency/` Is it possible to use test mode for the ResourceManager here? That way we only work with string IDs

src/taco-dependency-installer/test/javaJdkInstaller.ts
digeff: Done! Also added a default clause in the switch.

src/taco-dependency-installer/test/javaJdkInstaller.ts
digeff: Done!

src/taco-dependency-installer/test/javaJdkInstaller.ts
digeff: Yes, I forgot to do that :(

src/taco-dependency-installer/test/javaJdkInstaller.ts
guillaumejenkins: Nevermind this, I hadn't seen that there is a 4th test near the top of the suite that doesn't use this exec behavior. I instead suggested offline that the `fakeExecToTestPackageInstallerFailure` handler be moved to the top of this suite so that it's above all 4 tests.

guillaumejenkins: Nevermind this, I hadn't seen that there is a 4th test near the top of the suite that doesn't use this exec behavior. I instead suggested offline that the fakeExecToTestPackageInstallerFailure handler be moved to the top of this suite so that it's above all 4 tests.

digeff: Done!

PR #115, opened 2015-10-27
Changed calls to JSON.parse(fs.readFileSync()) to use 'utf8' so we don't need to cast the result of reading the file

digeff:
JSON.parse has the string type even though it accepts any object, so people will be forced to convert things to strings explicitly.


MSLaguana:
Looks good to me.

meenakb:
:+1:

PR #118, opened 2015-10-27
Combine external event name using "/" instead of "."

TimBarham:
I meant to include this change with my earlier PR.

For consistency with other event name (like those generated by dependency installer), this combines the external event name using "/" instead of "." (so for `taco simulate`, for example, we end up with event names like `TACO/simulate/exec` instead of `TACO/simulate.exec`.


MSLaguana:
Looks good to me.

PR #119, opened 2015-10-28
Final implementation of 'taco simulate' command.

TimBarham:
This includes the final command options, final help text, telemetry support and some tests.


package.json
```diff
@@ -7,6 +7,7 @@
     "description": "specifies dependencies needed for developing taco",
     "version": "0.0.1",
     "devDependencies": {
+        "taco-simulate": "^0.1.1",
```
MSLaguana: Why have you added this here? We don't need taco-simulate in order to compile/develop the packages, we'd just need it as a dependency of taco-cli surely?

TimBarham: Hmmm... That was something I added right at the very beginning and for the life of me I can't recall why. I'll remove.

src/taco-cli/test/simulate.ts
```diff
@@ -0,0 +1,113 @@
...
+                remain: args.slice()
+            };
+            simulate.run(commandData).then((telemetryParameters: tacoUtils.ICommandTelemetryProperties) => {
+                throw new Error("Scenario succeeded when it should have failed");
+            }, function (error: TacoError): void {
+                error.errorCode.should.equal(expectedErrorNumber);
+            }).done(() => done(), done);
+        }
+
+        it("returns an error when taco-simulate returns an error", function (done: MochaDone): void {
```
nisheetjain: we should add a case for scenario when no platform is added to the project

PR #120, opened 2015-10-28
Livereload/Devicesync integration with TACO

omefire:
This code is still evolving.
It's not meant to be pulled in in its current form. There's still work to be done on it.


omefire:
This enables the following command:
`taco run android –livereload`

If the user doesn’t have the plugin installed on his project, it gets installed and then the livereload session gets started.

src/taco-cli/cli/run.ts
```diff
@@ -186,7 +190,7 @@ class Run extends commands.TacoCommandBase {
             name: "local",
             run: Run.local,
             canHandleArgs(commandData: commands.ICommandData): boolean {
-                return !!commandData.options["local"];
+                return !!commandData.options["local"] || !!commandData.options["livereload"] || !!commandData.options["devicesync"];
```
MSLaguana: This might not have the desired effect: With the code this way, if I say `taco run ios` it will run it remotely if I have configured a remote, but `taco run ios --livereload` will try and run it locally. That seems a bit confusing to me; might be better to error out if livereload or devicesync are specified and we would otherwise run remotely?

src/taco-cli/cli/utils/cordovaWrapper.ts
```diff
@@ -170,10 +170,22 @@ class CordovaWrapper {
...
     }
 
-    public static run(commandData: commands.ICommandData, platforms: string[] = null): Q.Promise<any> {
-        return CordovaWrapper.cordovaApiOrProcess((cordova: Cordova.ICordova) => {
-            return cordova.raw.run(CordovaHelper.toCordovaRunArguments(commandData, platforms));
-        }, () => ["run"].concat(CordovaHelper.toCordovaCliArguments(commandData, platforms)));
+    public static run(commandData: commands.ICommandData, platforms: string[] = null): Q.Promise <any> {
+	return CordovaWrapper.cordovaApiOrProcess((cordova: Cordova.ICordova) => {
+            // First install the plugin if it's not yet installed on this project
+            return cordova.raw.plugin('ls', null, null).then(function(pluginList: string[]): Q.Promise <any> {
```
MSLaguana: This is going to always add the plugin, regardless of whether livereload or devicesync are specified. Perhaps it'd be better to move the plugin adding logic to the calling code? Or at least add a check for whether the flags are specified.

omefire: good catch. thanks.

src/taco-cli/cli/run.ts
omefire: The intent is to support livereload/devicesync only in local scenarios.
I thought the indicator of a command to be ran remotely was only the presence of the 'remote' keyword.
Is this the appropriate method to use to check for that ? : 

```IPlatformWithLocation.location === PlatformHelper.BuildLocationType.Local```

MSLaguana: If `--local` or `--remote` are specified then we only do local or remote targets, but if neither is specified then `taco run ios windows` will run both ios and windows, and will decide whether to build them remotely or not based on whether a remote has been configured. This isn't the right place to check that condition, since it is a per-platform value and isn't yet determined.

A better place to check would be runRemotePlatform which is where we try and actually run remotely; you could error out or ignore the options if they are specified there.

omefire: Agreed. this is not the right place to do it. I'm currently reworking the code to a better structure given what I've just learned.

omefire:
Closing this PR in favor of this one: https://github.com/Microsoft/TACO/pull/174

PR #121, opened 2015-10-28
Change tsconfig.json in TS template to match http://json.schemastore.…

rido-min:
tsconfig.json content should be {compilerOptions...} instead of { tsCompilerOptions.. } as specified in http://json.schemastore.org/tsconfig

### DIFF
--- a/templates/default/typescript/scripts/tsconfig.json
+++ b/templates/default/typescript/scripts/tsconfig.json
@@ -1,5 +1,5 @@
 {
-	"compileOptions" : {
+	"compilerOptions" : {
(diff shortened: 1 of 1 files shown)


nisheetjain:
@guillaumejenkins  can you review this?

MSLaguana:
I was just looking over this PR to see if we should merge it, but it looks like there are more references to tsCompileOptions in the code which might need to be modified as well. @guillaumejenkins can you look into this and verify that after we take the change everything still works? The other references might just be for our own compilation/gulp scripts.

PR #122, opened 2015-10-28
Adding support for Taco code Coverage 

nisheetjain:


src/remotebuild/package.json
```diff
@@ -42,12 +42,14 @@
...
         "mocha": "2.0.1",
+        "istanbul": "^0.4.0",
         "mkdirp": "0.3.5",
         "should": "4.3.0",
         "request": "2.36.0"
     },
     "scripts": {
-        "test": "mocha"
+        "test": "mocha",
+        "coverage": "istanbul cover --report json node_modules/mocha/bin/_mocha -- --reporter spec"
```
MSLaguana: Is there a reason you've used node_modules/mocha/bin/_mocha rather than node_modules/.bin/mocha ?

nisheetjain: yup on windows mocha ends up matching with mocha.cmd and istanbul fails to work with this. after lot of web search, found this is the right way to do this across platforms

MSLaguana: Ah ok, makes sense.

tools/GulpCoverageUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+        // read all the directories and prepare a flat list of coverageXXX.json we need to remap
+        return Q.all<string[]>(coverageResultsPaths.map(coverageResultsPath => {
+            coverageResultsPath = path.resolve(coverageResultsPath);
+            return Q.denodeify(fs.readdir)(coverageResultsPath)
+                .then(function(files: string[]): string[] {
+                    return files.filter(file => {
+                        return file.toLowerCase().indexOf("coverage-") === 0 && path.extname(file) === ".json";
+                    }).map(file => path.join(coverageResultsPath, file));
+                });
+        })).spread(function(...args: string[][]): string[] {
```
MSLaguana: Why spread and then use ...args? You can just use a .then taking the `string[][]` as its first argument

tools/GulpCoverageUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+                return coverageFilesList.concat(coverageFiles);
+            });
+
+        });
+    }
+
+    private static normalizeCoveragePaths(coverageJsonsList: string[], coverageOutputDir: string, modulesRoot: string): Q.Promise<string[]> {
+        var index: number = 0;
+        // put all extra coverage files in a random location under coverage
+        var randomDir: string = Math.floor(Math.random() * 10000) + 1000 + "";
```
MSLaguana: What's this for? we are hiding some files in a random place?

tools/GulpCoverageUtils.ts
nisheetjain: done

tools/GulpCoverageUtils.ts
nisheetjain: removed the random dir and replaced it with "secondary". something we delete post coverage

meenakb:
:+1:

PR #124, opened 2015-10-28
Fixes `CordovaWrapper.getGlobalCordovaVersion()` return value.

TimBarham:
This method uses `cordova -v` to get the current Cordova version. That command returns the version in the following format:

    5.3.3 (cordova-lib@5.3.3)

The existing logic wasn't stripping off the bit in brackets. As a result, calls to `semver` methods (such as `semver.gte()` in `CordovaWrapper.requirements()`) could throw an exception.


TimBarham:
@meenakb - does this look right to you?

meenakb:
:+1:  Yep!

PR #127, opened 2015-10-30
Allowing user configuration of app launching timeout

MSLaguana:
It seems that the time taken to communicate with the device can vary significantly between mac hosts, so I've bumped the default timeout and also allowed the end-user to specify the timeout if they desire.

### DIFF
--- a/src/taco-remote/lib/tacoRemoteConfig.ts
+++ b/src/taco-remote/lib/tacoRemoteConfig.ts
@@ -32,6 +32,8 @@ class TacoRemoteConfig implements RemoteBuild.IReadOnlyDictionary {
         webDebugProxyPortMin?: number;
         webDebugProxyPortMax?: number;
 
+        appLaunchStepTimeout?: number;
+
         [key: string]: any;
     };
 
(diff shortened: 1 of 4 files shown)


leotlee:
Maybe consider adding retries as well.  What was the worst case time you've seen it took to launch the app?

MSLaguana:
I'm not sure that retries at this level is appropriate; if the user wants to try again, they can send another run request (without doing another build, even) to retry.

As for the worst time, it's hard to tell, but the machine that was failing with a 5s timeout (On the second command sent, typically) seems to work reliably with a 10s timeout. It also looked like it was only barely failing the 5s timeout since the app would launch moments after it timed out.

nisheetjain:
:+1:

leotlee:
:+1:

src/taco-remote/lib/tacoRemoteConfig.ts
```diff
@@ -60,7 +63,6 @@ class TacoRemoteConfig implements RemoteBuild.IReadOnlyDictionary {
 
     /**
      * @name allowsEmulate
-     * @LOCTAG TacoRemoteConfigAllowsEmulate
```
meenakb: Are these LOCTAG deletions intentional?

MSLaguana: Yes, when I was adding the new property I noticed that these were unused and so I cleaned them up.

meenakb:
:+1:

PR #129, opened 2015-11-02
Fix remotebuild help

joshuaskelton:
## Summary
Added null checks in the HelpCommandBase class and added tests for all of the remotebuild help topics.

## Changes
- Null checks in HelpCommandBase.ts
- New tests in remotebuild/test/help.ts


src/remotebuild/package.json
```diff
@@ -34,6 +34,7 @@
         "q": "1.0.1",
         "rimraf": "2.2.6",
         "semver": "^4.3.6",
+		"colors": "^1.0.3",
```
MSLaguana: Looks like this is only used for a test, so I think it should be a devdependency

src/remotebuild/test/utils/memoryStream.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+*/
+
+import stream = require ("stream");
+
+import tacoUtils = require("taco-utils");
+
+import utils = tacoUtils.UtilHelper;
+
+export class MemoryStream extends stream.Writable {
```
MSLaguana: Rather than duplicating this class across packages, it might be a candidate for the taco-test-utils package

joshuaskelton: That is a good suggestion.

PR #130, opened 2015-11-02
Adding --failTestsAtEnd (--ftae) for "taco run-tests". Also adding

nisheetjain:
shorthands for other TACO gulp tasks


MSLaguana:
Looks good to me.

How does moduleFilter work? Suppose I ran tests and was told that taco-remote and taco-utils had a test failure, can I do something like `gulp run-tests --modulefilter taco-utils --modulefilter taco-remote` and have it just rerun those two?

nisheetjain:
@MSLaguana  yes you can. you can use "gulp run-tests --ftae --m taco-utils --m taco-remote"

PR #133, opened 2015-11-03
Removing gulp prep/unprep tasks. New corresponding steps would "npm i…

nisheetjain:
Removing gulp prep/unprep tasks. New corresponding steps would "npm install" and "npm

run clean" which should be run from TACO root folder


nisheetjain:
thoughts?

MSLaguana:
Looks good to me. Does "npm install" and "npm run clean" have to happen at the root, or can you do it from within the package (e.g. src folder)? I'm wondering if we should explicitly use `__dirname` to make paths absolute if that is the case.

PR #134, opened 2015-11-03
Exposing mocha reporter to gulp. this is needed to make sure we can run

nisheetjain:
gulp run-tests with a custom reporter


MSLaguana:
Looks good to me

PR #138, opened 2015-11-04
Fix remotebuild help crash and add tests

joshuaskelton:
## Summary
Added null guards to the HelpCommandBase when we resolve aliases. Added tests for remotebuild help, and some test refactoring.

## Changes
* Added null checks to HelpCommandBase class when resolving aliases.
* Added tests for remotebuild help commands.
* Moved the MemoryStream class to the tacoTestUtils module.
* Added typings for MemoryStream.
* Updated existing tests that used MemoryStream

### DIFF
--- a/src/remotebuild/package.json
+++ b/src/remotebuild/package.json
@@ -44,7 +44,8 @@
         "mocha": "2.0.1",
         "mkdirp": "0.3.5",
         "should": "4.3.0",
-        "request": "2.36.0"
+        "request": "2.36.0",
+        "colors": "^1.0.3"
     },
     "scripts": {
         "test": "mocha"
--- a/src/remotebuild/package.json
+++ b/src/remotebuild/package.json
@@ -45,7 +45,8 @@
         "mkdirp": "0.3.5",
         "should": "4.3.0",
         "request": "2.36.0",
-        "colors": "^1.0.3"
+        "colors": "^1.0.3",
+        "taco-tests-utils": "file:../taco-tests-utils"
     },
     "scripts": {
         "test": "mocha"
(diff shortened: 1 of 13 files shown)


src/remotebuild/test/help.ts
```diff
@@ -0,0 +1,253 @@
...
+import tacoTestsUtils = require("taco-tests-utils");
+
+import UtilHelper = tacoUtils.UtilHelper;
+import ms = tacoTestsUtils;
+
+import ICommandData = tacoUtils.Commands.ICommandData;
+
+function parseRemoteBuildConf(): RemoteBuildConf {
+    // Configuration preference: command line, then anything in a config file if specified by the --config arg, then some defaults for options not yet set
+    nconf.argv();
```
MSLaguana: Is this appropriate for the tests? If we aren't going to mess around with process.argv then perhaps it would be better to go straight to using a config file, or nconf.overrides or something like that to specify particular arguments?

src/remotebuild/package.json
```diff
@@ -44,7 +44,8 @@
         "mocha": "2.0.1",
         "mkdirp": "0.3.5",
         "should": "4.3.0",
-        "request": "2.36.0"
+        "request": "2.36.0",
+        "colors": "^1.0.3"
```
MSLaguana: You'll need to add taco-test-utils to the devdependencies too

src/remotebuild/test/help.ts
joshuaskelton: Turns out I don't need a special config, so I've mostly removed this code.

src/remotebuild/package.json
joshuaskelton: Good catch. Done.

MSLaguana:
Looks good to me

PR #141, opened 2015-11-05
Merge 1.1.0 back to master

MSLaguana:
Taking the hotfixes made in the 1.1.0 branch back into master.


src/taco-cli/cli/utils/cordovaHelper.ts
```diff
@@ -280,12 +281,19 @@ class CordovaHelper {
         });
     }
 
+    public static ensureCordovaVersionAcceptable(cliVersion: string): void {
+        if (semver.valid(cliVersion) && semver.lt(cliVersion, "5.4.0") && semver.gte(process.versions.node, "5.0.0")) {
```
joshuaskelton: There is similar logic in taco-remote-lib/common/builder.ts. Wouldn't it be better to have a single source of truth for checking the known good range of Cordova versions?

MSLaguana: We discussed that when we were making the fix initially. taco-remote-lib currently has 2 version checks, one equivalent to this one here, and one checking for node 4.X and cordova < 5.3.3. That second check only applies to iOS and isn't as relevant to taco-cli, unlike this check. We hope that there won't be many more of these sudden incompatibilities that we can't work around, so to try and avoid over engineering we've gone with just putting in the checks here for now.

If this type of thing happens again, then we'll definitely go and work out a better architecture so we don't need to have manually-maintained mappings of good and bad versions stretching on forever.

joshuaskelton: :+1:

PR #142, opened 2015-11-06
Move old AndroidSdkInstaller tests that were related to the package installer, to the AndroidPackagesInstaller tests, and fixed them

digeff:
- Moved skiped AndroidSdkInstaller tests to a new file androidPackagesInstaller.ts
- Fixed the tests
- Refactored some AndroidPackagesInstaller code

Manual tests:
  * Run install-reqs with no Android SDK installed
  * Run install-reqs when only the target is missing
  * Run install-reqs when android.bat doesn't exit
  * Run install-reqs when android.bat exits with an error code
  * Run install-reqs when adb.exe doesn't exist (I removed it after it was installed)

### DIFF
--- a/src/taco-dependency-installer/installers/androidPackagesInstaller.ts
+++ b/src/taco-dependency-installer/installers/androidPackagesInstaller.ts
@@ -77,26 +77,26 @@ class AndroidPackagesInstaller extends InstallerBase {
         }
     }
 
-    private getAvailableAndroidPackages(): Q.Promise<any> {
-        var deferred: Q.Deferred<any> = Q.defer();
+    private executeAndroidCommand(args: string[], nameForTelemetry: string,
+        shouldAutomaticallyAcceptLicenseTerms: boolean): Q.Promise<string> {
+        var deferred: Q.Deferred<string> = Q.defer<string>();
+        var stdErrOutput: string = "";
+        var stdOutput: string = "";
 
-        // Call "android list sdk -e" and capture output
-        var args: string[] = [
-            "list",
-            "sdk",
-            "-e"
-        ];
-        var output: string = "";
-        var errorOutput: string = "";
         var cp: childProcess.ChildProcess = os.platform() === "darwin" ?
             childProcess.spawn(this.androidCommand, args, { uid: parseInt(process.env.SUDO_UID, 10), gid: parseInt(process.env.SUDO_GID, 10) }) : childProcess.spawn(this.androidCommand, args);
 
         cp.stdout.on("data", function (data: Buffer): void {
-            output += data.toString();
+            stdOutput += data.toString();
         });
+
+        cp.stderr.on("data", function (data: Buffer): void {
+            stdErrOutput += data.toString();
+        });
+
         cp.on("error", (err: any) => {
             this.telemetry
-                .add("error.description", "ErrorOnChildProcess on installAndroidPackages", /*isPii*/ false)
+                .add("error.description", "ErrorOnChildProcess on " + nameForTelemetry, /*isPii*/ false)
                 .addError(err);
 
             if (err.code === "ENOENT") {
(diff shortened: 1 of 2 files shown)


src/taco-dependency-installer/installers/androidPackagesInstaller.ts
```diff
@@ -77,26 +82,25 @@ class AndroidPackagesInstaller extends InstallerBase {
         }
     }
 
-    private getAvailableAndroidPackages(): Q.Promise<any> {
-        var deferred: Q.Deferred<any> = Q.defer();
+    private executeAndroidCommand(name: string, args: string[]): IExecuteCommandResult {
```
guillaumejenkins: I think it would be more intuitive to just return a promise resolved with stdout... I think it's slightly confusing to have to use this method by chaining to `result.output` like such:

```typescript
var cp: IExecuteCommandResult = this.executeAndroidCommand("getAvailableAndroidPackages", args);
return cp.output.then((output: string) => {
```

I understand that you had to do it this way for the `installAndroidPackages` method to be able to listen for the "[y/n]:" prompt. I understand that we want `executeAndroidCommand` to be as generic as possible, but I still feel it would make sense to move the "[y/n]:" prompt logic to `executeAndroidCommand`. This way, consumers that want to invoke an Android command don't need to worry about any prompt, and we get to change the signature to just return a promise for stdout. We can then use this method like so:

```typescript
this.executeAndroidCommand("getAvailableAndroidPackages", args).then((output: string) => {
...
});
```

which I believe is much more intuitive. But it's just my opinion, feel free to disagree!

guillaumejenkins: Also, small nit: it took me a few moments to realize that the `name` parameter was just for telemetry purposes. I think it would be good to make it more obvious. I would put it as second parameter, and make it optional (default to args[0]). That way, consumers can invoke Android commands without worrying about telemetry-specific stuff. I would also change the name slightly to reflect better that it's a telemetry param.

So, this is what I propose (combined with my above comment):
```typescript
private executeAndroidCommand(args: string[], telemetryStep?: string): Q.Promise<string> {
    if(!telemetryStep) {
        telemetryStep = args[0]; // You may have to deal with null / empty args?
    }
    // ...
}
```

digeff: If we are just going to use this executeAndroidCommand method for these two commands, I think it'd be okay to move the y/n logic inside the command and use a boolean parameter to trigger it.

If we are going to have additional commands in the future, and they are not going to need this y/n logic, I think I'd be best to leave it out of this method.

Do you know if we have any plans to update this code, and add additional calls to the android commands?

digeff: Regarding the second comment, what do you think of this signature?:
executeAndroidCommand(args: string[], nameForTelemetry: string);

telemetryStep has the connotation of telemetry.step(); and we are not doing that, so I'd like to use a different concept.
Regarding the optional value, the telemetry name is quite important for us to use telemetry effectively, and I think that whoever calls this should spend some time thinking of a good name for this, so I want it to be mandatory.

guillaumejenkins: First comment: I can't think of any other Android command we would need other than list and update. But even if we did add new methods, and they don't involve a prompt, I still think it's good to have the low-level details of how to bypass a prompt hidden in the execute command. Can we instead have a parameter `expectPrompt: boolean = false` to trigger this logic, and another `acceptPrompt: boolean = true`? This way, the callers are still in control, but we don't have to duplicate the prompt logic in every caller

Second command: I like your new name and I also agree to leave it mandatory

guillaumejenkins: Other than that :+1:

digeff: I changed the signature to:
executeAndroidCommand(args: string[], nameForTelemetry: string,
        shouldAutomaticallyAcceptLicenseTerms: boolean): Q.Promise<string>
And move the accepting license terms logic inside given that we only have two callers, I think it's clearer to have all parameters mandatory.

What are you referring to with: expectPrompt: boolean?

guillaumejenkins: It was just in case we wanted the callers to let the execute command know whether there is going to be a prompt or not, but it's not really needed.

I think this is good as is :+1:

PR #144, opened 2015-11-09
Refactoring selftest and adding check for plugins

MSLaguana:
Also bumping cordova version in selftest for maximum nodejs compatibility


nisheetjain:
:+1:

PR #145, opened 2015-11-09
Adding in error handling for missing homebrew packages

MSLaguana:


src/taco-remote-lib/ios/ios.ts
```diff
@@ -218,6 +218,15 @@ class IOSAgent implements ITargetPlatform {
                 res.status(200).json(buildInfo.localize(req, resources));
             }
         });
+
+        ideviceinstaller.on("error", function (err: any): void {
+            res.status(500);
+            if (err.code === "ENOENT") {
+                res.send(resources.getStringForLanguage(req, "IdeviceinstallerNotFound"));
```
nisheetjain: Can you fix this casing as well? looks like something I regressed during camelCasing conversions

MSLaguana: Which casing are you referring to? The string ID here is new, and if you prefer a different casing then I can change it.

nisheetjain: this should be IDeviceXXX right?

MSLaguana: The name of the tool is "ideviceinstaller", so I went with "IdeviceinstallerNotFound", but it's easy enough to change. This isn't something that you broke earlier on, don't worry about that :)

nisheetjain:
:+1:  looks good otherwise

PR #149, opened 2015-11-12
Increasing the range for dependency installer error codes and fixing a

nisheetjain:
test fix where wrong range is being used

@meenakb , @guillaumejenkins can you please take a look

### DIFF
--- a/src/taco-cli/package.json
+++ b/src/taco-cli/package.json
@@ -42,7 +42,6 @@
         "rimraf": "2.2.6",
         "semver": "^4.3.6",
         "taco-dependency-installer": "file:../taco-dependency-installer",
-        "taco-tests-utils": "file:../taco-tests-utils",
         "taco-utils": "file:../taco-utils",
         "wrench": "^1.5.8",
         "colors": "^1.0.3",
@@ -56,7 +55,8 @@
         "gulp-concat": "^2.4.2",
         "mocha": "latest",
         "istanbul": "^0.4.0",
-        "should": "4.3.0"
+        "should": "4.3.0",
+        "taco-tests-utils": "file:../taco-tests-utils"
     },
     "readmeFilename": "README.md",
     "scripts": {
--- a/src/taco-cli/package.json
+++ b/src/taco-cli/package.json
@@ -42,6 +42,7 @@
         "rimraf": "2.2.6",
         "semver": "^4.3.6",
         "taco-dependency-installer": "file:../taco-dependency-installer",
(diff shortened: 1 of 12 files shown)


meenakb:
:+1:

joshuaskelton:
:+1:

src/taco-dependency-installer/test/tacoErrors.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/// <reference path="../../typings/mocha.d.ts" />
+
+"use strict";
+
+import mocha = require ("mocha");
+import path = require ("path");
+
+import resources = require ("../resources/resourceManager");
+import testUtil = require ("../../taco-utils/testUtils/testHelper");
```
guillaumejenkins: Are we sure we want to require cross-package files using relative paths like this? When this is installed via NPM install, taco-utils most likely won't be in the same root dir as taco-dependency-installer... It will be under ./node_modules/taco-utils, so I believe this will make these tests fail

MSLaguana: Agreed. It would be better to make sure that taco-utils is a dependency in package.json, and then `require("taco-utils/testUtils/testHelper")` without the relative path to refer to the particular file. Not sure whether that will provide intellisense/typings appropriately though.

nisheetjain: good point @guillaumejenkins  I copied an existing tacoError test. let me fix both

src/taco-cli/package.json
```diff
@@ -42,6 +42,7 @@
         "rimraf": "2.2.6",
         "semver": "^4.3.6",
         "taco-dependency-installer": "file:../taco-dependency-installer",
+        "taco-tests-utils": "file:../taco-tests-utils",
```
MSLaguana: Should this be a devDependency?

src/taco-dependency-installer/package.json
```diff
@@ -26,6 +26,7 @@
     "q": "^1.1.2",
     "request": "2.36.0",
     "taco-utils": "file:../taco-utils",
+    "taco-tests-utils": "file:../taco-tests-utils",
```
MSLaguana: This should also be a devDependency

src/taco-cli/package.json
guillaumejenkins: Also we should double check that it is correctly picked up by updateDynamicDependencies.js to make sure these tests don't fail if we build the sources on one computer, transfer them to a new computer and then try to run the tests

PR #150, opened 2015-11-13
Adding a gulp link to allow "sudo gulp link" on non windows machines

nisheetjain:


meenakb:
:+1:

tools/gulpmain.ts
```diff
@@ -64,7 +64,14 @@ if (options.modulesFilter) {
 }
 
 /* Default task for building /src folder into /bin */
-gulp.task("default", ["install-build"]);
+gulp.task("default", [], function (callback: gulp.TaskCallback): void {
+    var os = require ("os");
+    if (os.platform() === "win32") {
+        runSequence("install-build", "link", callback);
```
MSLaguana: This could still fail if someone has their npm prefix pointing to program files, which I've seen sometimes. Not as common though.

nisheetjain: let's unblock CI server for now. we can discuss this further :)

PR #152, opened 2015-11-13
Introducing error levels for TacoError

meenakb:


src/taco-utils/tacoError.ts
```diff
@@ -36,12 +42,23 @@ module TacoUtility {
...
 
         public get isTacoError(): boolean {
             return true;
         }
 
+        public static getWarning(errorToken: string, resources: ResourceManager, ...optionalArgs: any[]): TacoError {
+            var message: string = TacoError.getMessageString(errorToken, resources, optionalArgs);
+
+            // We do not use an error code for Warnings
+            var warning = new TacoError (-1, message);
```
nisheetjain: let's use 0 instead of -1

nisheetjain:
:+1:

PR #154, opened 2015-11-13
Adding in support for 'taco run --list'

MSLaguana:


src/taco-cli/cli/utils/cordovaHelper.ts
```diff
@@ -159,6 +159,13 @@ class CordovaHelper {
         return CordovaHelper.toCordovaArgumentsInternal(commandData, platforms);
     }
 
+    public static toCordovaTargetsArguments(commandData: commands.ICommandData, platforms: string[] = null): Cordova.ICordovaRawOptions {
+        // Run, build, emulate, prepare and compile all use the same format at the moment
+        return CordovaHelper.toCordovaArgumentsInternal(commandData, platforms);
```
dlebu: Do we really need separate implementations if they are the same? Is this going to change in the future based on the command?

src/taco-cli/resources/en/resources.json
```diff
@@ -378,6 +378,8 @@
     "_CommandRunEmulatorDescription.comment": "Description of the --emualtor argument to 'taco run'",
     "CommandRunTargetDescription": "Run a build on the specified target",
     "_CommandRunTargetDescription.comment": "Description of the --target argument to 'taco run'",
+    "CommandRunListDescription": "Report possible targets to run on without actually running",
```
dlebu: "List the available run targets without...." ?

dlebu:
Looks good!

src/taco-cli/cli/utils/cordovaHelper.ts
MSLaguana: Different commands may care about different sets of arguments, but at the moment they do go through a common place in the cordova code so we may be overengineering.

omefire:
Looks Good To Me!

src/taco-cli/resources/en/resources.json
nisheetjain: +1. how about we use same description as cordova, "List available targets". do we really need "without.."

omefire: +1 to that.

src/taco-cli/cli/commands.json
```diff
@@ -399,6 +399,9 @@
             ,
             {
                 "name": "--target=TARGET", "description": "[CommandRunTargetDescription]"
+            },
+            {
+                "name":  "--list", "description": "[CommandRunListDescription]"
```
meenakb: Cordova 5.4 provides a DEVICECLASS option for users to specify if they want to list device or emulator targets. You should add that to the command's help description.

src/taco-cli/resources/en/resources.json
meenakb: Additionally, specify that both both device and emulator targets will be listed (unless DEVICECLASS option is specified).

src/taco-cli/cli/commands.json
nisheetjain: we should definitely list that as an option to the command/synopsis. Description, not so sure because it will make the string very long...

MSLaguana: The option is already there (--device and --emulate)

Cordova lays its help out in a different way to us, with the options grouped into "bags" such as the DEVICECLASS bag which is --device or --emulator, the PROD bag of --debug or --release, etc.

meenakb: Yes, let us atleast be explicit that when not specifying the device/emulator options in conjunction with the --list option, both targets will be listed.

PR #155, opened 2015-11-13
Modifying 'taco platform list' to include remote platforms

MSLaguana:


src/taco-cli/cli/utils/platformPluginCommandBase.ts
```diff
@@ -146,6 +153,9 @@ export class PlatformPluginCommandBase extends commands.TacoCommandBase {
             return cordovaWrapper.invokePlatformPluginCommand(self.name, self.cordovaCommandParams, data);
         })
             .then(function (): Q.Promise<any> {
+            return self.postInvokeStep(data);
```
nisheetjain: not sure if I follow, why do we need to execute add/remove related code like specsToPersist for "list" command? can we use existing pattern like "run" and invoke list via CordovaWrapper?

src/taco-cli/cli/platform.ts
```diff
@@ -127,13 +129,24 @@ class Platform extends commandBase.PlatformPluginCommandBase {
...
 
             case CommandOperationStatus.Success: {
-                logger.logLine();
                 this.printSuccessMessage(platforms, operation);
                 break;
             }
         }
     }
 
+    public postInvokeStep(data: commands.ICommandData): Q.Promise<any> {
```
meenakb: How about a better name for this method?

meenakb:
:+1:  otherwise

src/taco-cli/cli/utils/platformPluginCommandBase.ts
MSLaguana: Currently all "taco platform X" and "taco plugin X" commands go through this class doing the same steps, eventually deferring to corodvaWraper.invokePlatformPluginCommand just above. If we want to refactor these classes to use the subcommand style that the other commands use, it'll be a bigger task.

nisheetjain: can we do something like this in run module? 
    public run(data: commands.ICommandData): Q.Promise<ICommandTelemetryProperties> {
        if (data.options.remain == "--list") {
            // run list command logic 
        } else {
        return super.run(data);
        }
    }

src/taco-cli/cli/platform.ts
joshuaskelton: :+1:

PR #158, opened 2015-11-17
Updating taco-remote-lib to cope with cordova 5.4.0

MSLaguana:
Cordova 5.4.0 uses a new option type to pass arguments, and this change adapts to that. taco-remote-multiplexer now inspects the cordova version, and selects between a new version and the previous version depending on which cordova is targeted.


nisheetjain:
Looks good to me. yes I agree with the change. This is how we originally planned it :)

PR #159, opened 2015-11-17
Telemetry prompt at startup and on 'taco feedback' can now accept stream from stdin

meenakb:


src/taco-utils/telemetry.ts
```diff
@@ -159,11 +159,12 @@ module TacoUtility {
 
             var promptStringId: string = currentOptIn ? "TelemetryCurrentlyOptedInPrompt" : "TelemetryCurrentlyOptedOutPrompt";
 
-            newOptIn = TelemetryUtils.getUserConsentForTelemetry(utilResources.getString(promptStringId, Telemetry.appName));
-
-            // Change and save the new setting
-            TelemetryUtils.setTelemetryOptInSetting(newOptIn);
-            Telemetry.isOptedIn = newOptIn;
+            return TelemetryUtils.getUserConsentForTelemetry(utilResources.getString(promptStringId, Telemetry.appName))
+            .then( function(userOptedIn: boolean) {
```
MSLaguana: This function needs a return type

src/taco-utils/telemetry.ts
```diff
@@ -280,29 +289,48 @@ module TacoUtility {
...
                 logger.logLine();
-                var readlineSync: any = require("readline-sync");
-                return !!readlineSync.keyInYNStrict(LogFormatHelper.toFormattedString(optinMessage));
+
+                var deferred: Q.Deferred<boolean> = Q.defer<boolean>();
+                var yesOrNoHandler = readline.createInterface({ input: process.stdin, output: process.stdout });
+
+                yesOrNoHandler.question(LogFormatHelper.toFormattedString(optinMessage), function (answer: string): void {
+                    yesOrNoHandler.close();
+                    if (answer && utilResources.getString("PromptResponseYes").toLowerCase().split("\n").indexOf(answer.trim()) !== -1) {
```
MSLaguana: This approach is better for localization too :)

src/taco-utils/telemetry.ts
```diff
@@ -280,29 +289,48 @@ module TacoUtility {
...
-                    optIn = TelemetryUtils.getUserConsentForTelemetry(utilResources.getString("TelemetryOptInQuestion"));
-                    TelemetryUtils.setTelemetryOptInSetting(optIn);
+                    TelemetryUtils.getUserConsentForTelemetry(utilResources.getString("TelemetryOptInQuestion"))
+                    .then( function(userOptedIn: boolean) {
+                        optIn = userOptedIn;
+                        TelemetryUtils.setTelemetryOptInSetting(optIn);
+                        deferred.resolve(optIn);
+                    });
+                } else {
+                    deferred.resolve(optIn);
```
MSLaguana: No need to use a deferred here; you can just do
```
if (_.isUndefined(optIn)) {
    [...] 
    return TelemetryUtils.getUserConsentForTelemetry([...])
    .then( function (theValue) {
       [...]
       return theValue
    )}
} else {
    return Q(optIn);
}
```

MSLaguana:
Looks good to me! So with this we can now `echo yes | taco feedback` or `echo no | taco feedback` and everything works out appropriately, right?

meenakb:
Yep, that is the intent of the change :)

PR #160, opened 2015-11-17
Fixing CI tests

MSLaguana:
Increasing the timeout on the taco-remote selftest, since we how include a plugin in the build which is acquired from the internet.

Also fixing a failure I introduced in the taco help tests, as well as improving the tests to show up failures as failures rather than timeouts.


nisheetjain:
looks good...

MSLaguana:
This fixes a couple more issues, but there's still a timeout in the kit tests that I haven't tracked down yet.

PR #161, opened 2015-11-18
Fix to allow taco.json which is not UTF-8 formatted. Fix supports UTF-8

nisheetjain:
with BOM characters and UTF-16. TACO can use parseUserJSON going forward
to parse JSONs which are exposed to end user


src/taco-utils/utilHelper.ts
```diff
@@ -320,6 +320,36 @@ module TacoUtility {
...
+            var contents: Buffer = fs.readFileSync(filePath);
+            // try the simplest path first
+            try {
+                return JSON.parse(contents.toString());
+            } catch (ex) {
+                UtilHelper.emptyMethod();
+            }
+
+            // may be this is a UTF-8 with BOM
+            var iconv = require("iconv-lite");
```
MSLaguana: Why require iconv here rather than up the top with everything else?

nisheetjain: delay load?

MSLaguana: This way doesn't get the typing information though. It's not a huge deal.

src/taco-utils/utilHelper.ts
```diff
@@ -320,6 +320,36 @@ module TacoUtility {
...
+                return JSON.parse(iconv.decode(contents, "utf8"));
+            } catch (ex) {
+                UtilHelper.emptyMethod();
+            }
+            // May be this is a UTF-16 file
+            try {
+                return JSON.parse(iconv.decode(contents, "utf16"));
+            } catch (ex) {
+                UtilHelper.emptyMethod();
+            }
```
digeff: I think we should throw an exception if we can't parse the file...

nisheetjain: I agree, let me add that..

meenakb:
Changes look good. Can you please use this util method for parsing of all JSON files in our codebase? Espcially, TacoKitMetadata.json. I see that with kit extensiblity in future, this would provide value.

nisheetjain:
API's name is parseUserJSON because it's meant for json files which can get modified by user. for for TacoKitMetadata.json we control the encoding. but good point, if we support extensible kits, we should use parseUserJSON to parse those kit files.

nisheetjain:
@digeff @MSLaguana looks good?

src/taco-utils/test/utilHelper.ts
```diff
@@ -163,4 +173,70 @@ describe("UtilHelper", function (): void {
...
+        });
+
+        it("read UTF-8 with BOM", function (): void {
+            var parsedJson = UtilHelper.parseUserJSON(path.resolve(__dirname, "taco_utf8BOM.json"));
+            parsedJson["cordova-cli"].should.equal("5.3.3");
+        });
+
+        it("read file which doesn't exist", function(): void {
+            try {
+                UtilHelper.parseUserJSON(path.resolve(__dirname, "nonExistentFile.json"));
```
digeff: If this doesn't throw an exception the test will pass anyway.

src/taco-utils/test/utilHelper.ts
```diff
@@ -163,4 +173,70 @@ describe("UtilHelper", function (): void {
...
+
+        function writeBufferAndValidate(filename: string, getBuffer: (stringified: string) => Buffer, expectedErrorCode?: TacoErrorCodes) {
+            var x = { "cordova-cli": 5.8 };
+            var stringified: string = JSON.stringify(x);
+            var buffer: Buffer = getBuffer(stringified);
+
+            var filepath = path.join(testHome, filename);
+            fs.writeFileSync(filepath, buffer);
+            try {
+                var parsedJson = UtilHelper.parseUserJSON(filepath);
```
digeff: Same issue here. If expectedErrorCode is true, but this doesn't throw an exception, the validation will still pass

src/taco-utils/test/utilHelper.ts
```diff
@@ -163,4 +173,70 @@ describe("UtilHelper", function (): void {
...
+                if (!expectedErrorCode) {
+                    JSON.stringify(parsedJson).should.equal(stringified);
```
digeff: I think that should.equal also works properly for JSON objects, so you might be able to do parsedJson.should.equal(x)

src/taco-utils/utilHelper.ts
```diff
@@ -320,6 +322,39 @@ module TacoUtility {
...
+                }
+                // May be this is a UTF-16 file
+                try {
+                    return JSON.parse(iconv.decode(contents, "utf16"));
+                } catch (ex) {
+                    UtilHelper.emptyMethod();
+                }
+            }
+
+            throw errorHelper.get(tacoErrorCodes.TacoErrorCode.ErrorUserJsonMissingOrMalformed, filePath);
```
digeff: Would it make sense to split this into 2 errors? One for file missing and one for malformed?

nisheetjain: I don't see much added value with that. "missing or malformed" tells clearly either the file doesn't exist or is missing

src/taco-utils/test/taco_utf8BOM.json
```diff
@@ -0,0 +1,3 @@
+﻿{
```
MSLaguana: I assume that this has a BOM that just isn't showing up in the github preview?

src/taco-utils/test/utilHelper.ts
MSLaguana: I believe you'd want `.eql` then which does deep equality, rather than a shallow `==`

src/taco-utils/test/utilHelper.ts
```diff
@@ -163,4 +173,70 @@ describe("UtilHelper", function (): void {
...
+        after(function (): void {
+            rimraf(testHome, function (err: Error): void {/* ignored */ }); // Not sync, and ignore errors
+        });
+
+        it("read UTF-16 file", function (): void {
+            writeBufferAndValidate("utf16.json", (stringified: string) => new Buffer(stringified, "utf16le"));
+        });
+
+        it("read UTF-8 file", function (): void {
+            writeBufferAndValidate("utf8.json", stringified => new Buffer(stringified, "utf8"));
```
MSLaguana: Anonymous function should have a type

MSLaguana:
Looks good once you address @digeff's comments about false-successes.

PR #162, opened 2015-11-19
Fixing some formatting issues with TacoError

MSLaguana:
There was a flaw in how TacoError extracted arguments meaning that errors often had incorrect information in them. This change fixes that error, corrects how variadic arguments are passed through, and adds in a test case to make sure that we catch future regressions.


nisheetjain:
looks good to me

joshuaskelton:
:+1:

PR #163, opened 2015-11-19
Fixing build break caused by commit bd6940e. Commit introduced a cyclic

nisheetjain:
dependency.
Fix includes taco-tests-utils taking a proper dependency on taco-utils
and taco-utils removing it's dependency on taco-tests-utils

### DIFF
--- a/src/taco-tests-utils/package.json
+++ b/src/taco-tests-utils/package.json
@@ -20,6 +20,7 @@
     "lodash": "^3.10.1",
     "q": "~1.4.1",
     "mock-spawn": "~0.2.6",
+    "taco-utils": "file:../taco-utils",
     "should": "4.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


nisheetjain:
@digeff can you please take a look at changes to telemetryFakes?

digeff:
Is there any way to avoid duplicating the code?
Would declaring taco-tests-utils as a devDependency help? Or trying to include the file directly with a .. path?

MSLaguana:
Including a cross-package file via a relative path is a bad idea, since we do want these tests to work outside of our dev environment idealy. Making taco-tests-utils a devDependency would work except we then have a cyclic dependency between taco-utils and taco-tests-utils which is why it was removed in this PR.

We could try to split up taco-utils into a more granular structure and see if we can remove the cycle, or we could combine them both together so the cycle collapses to a single package.

nisheetjain:
@digeff  we have to make a call. either taco-tests-utils take dependency on taco-utils or the other way around. We can't have cyclic dependency.
Now since telemetryGenerateBase is a harder to break product dependency, I feel we should keep as taco-utils as the base package of everything. that we don't keep any test specific classes in utils (and don't expose them in package)

PR #166, opened 2015-11-19
Refactoring cordovaHelper, cordovaWrapper and projectHelper into

nisheetjain:
taco-utils. This allows other TACO packages to leverage APIs exposed in
these modules
Key Changes include:
1. Splitting ProjectHelper.editTacoJsonFile into
KitHelper.editTacoJsonFile to allow kit specific logic in taco-cli
2. Added VerifyExcludedTacoErrors to ensure all moved taco Errors are not
taken for some other values


nisheetjain:
@meenakb can you review this?

src/taco-cli/cli/tacoErrorCodes.ts
```diff
@@ -4,9 +4,6 @@
 // Error Codes: 5000- 5999
 enum TacoErrorCode {
     // General errors 5000-5449
-    CordovaCmdNotFound = 5001,
```
MSLaguana: I would leave these errors here, with comments indicating that they have been moved (or maybe change them to _CordovaCmdNotFound so we don't accidentally autocomplete to them)

Otherwise it'll be very hard to not reuse them, especially things like 5015 and 5016 which were at the end of their block

src/taco-cli/cli/build.ts
```diff
@@ -34,7 +32,10 @@ import TacoErrorCodes = require ("./tacoErrorCodes");
 import tacoUtility = require ("taco-utils");
 
 import commands = tacoUtility.Commands;
+import CordovaWrapper = tacoUtility.CordovaWrapper;
+import CordovaHelper = tacoUtility.CordovaHelper;
```
meenakb: CordovaHelper is not used in this class

src/taco-cli/test/tacoErrors.ts
```diff
@@ -23,5 +23,16 @@ describe("Taco Errors in TACO", function (): void {
     it("Verify TACO Errors in TACO", function (): void {
         TacoErrorTestHelper.verifyTacoErrors(path.join(__dirname, "../cli/tacoErrorCodes.js"), resources, 5000, 5999);
     });
+    it("Verify removed TACO Errors are not taken", function (): void {
+        TacoErrorTestHelper.verifyExcludedTacoErrors(path.join(__dirname, "../cli/tacoErrorCodes.js"), resources, 5001);
```
MSLaguana: Why not accept an array rather than call it once for each number?

src/taco-cli/cli/tacoErrorCodes.ts
meenakb: +1 - Another alternative is to just leave them in comments

src/taco-tests-utils/testProjectHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+import os = require ("os");
+import path = require ("path");
+import Q = require ("q");
+
+module TacoTestsUtils {
+    interface IKeyValuePair<T> {
+        [key: string]: T;
+    }
+
+    export class ProjectHelper {
```
meenakb: Can we just mock this class instead of duplicating?

src/taco-utils/resources/en/resources.json
```diff
@@ -76,5 +76,22 @@
...
+    "CommandCreateTacoJsonFileWriteError": "Failed to write to {0}",
+    "CommandCreateOnlyLocalCustomWww": "Error: Only local paths for custom www assets are supported.",
+    "_CommandCreateOnlyLocalCustomWww.comment": "Error reported when the specified path to the --copy-from or --link-to parameters is a web URL.",
+    "InvalidCordovaWithNode5": "Incompatible configuration: Unable to use Cordova < 5.4.0 with Node >= 5.0.0. Either upgrade to a newer version of Cordova, or revert to an older version of node. See http://go.microsoft.com/fwlink/?LinkID=618471 for more details.",
+    "_InvalidCordovaWithNode5.comment": "Error message printed when trying to use Cordova < 5.4.0 with Node >= 5.0.0.",
+    "CommandInstallCordovaTooOld": "Error: The current project uses Cordova version {0}. 'taco install-reqs' supports projects using Cordova version {1} or higher.<br/><br/>If your project was created using a Kit, consider moving to a kit that has a more recent version of Cordova. If your project was created using Cordova directly, consider updating the installation of Cordova. For more information, please refer to http://go.microsoft.com/fwlink/?LinkId=618544.",
```
meenakb: This error feels too TACO CLI specific to be a taco-utils candidate. Maybe get the relevant command as a param?

src/taco-tests-utils/tacoErrorTestHelper.ts
```diff
@@ -38,4 +34,20 @@ export class TacoErrorTestHelper {
             }
         });
     }
+
+    public static verifyExcludedTacoErrors(fileName: string, resources: any, excludedErrorCode: number): void {
+        var errorCodes: any = TacoErrorTestHelper.getErrorCodes(fileName);
+        var excludedErrorCodeStr: string = errorCodes[excludedErrorCode];
+        shouldModule(excludedErrorCodeStr).equal(undefined, "Exclude error code " + excludedErrorCode + "shouldn't be present");
```
MSLaguana: It might be slightly clearer to do `shouldModule(excludedErrorCode in errorCodes).be.false` to make sure that excludedErrorCode isn't a key, and not just a key mapping to undefined? Shouldn't really matter in our case though.

src/taco-utils/projectHelper.ts
```diff
@@ -0,0 +1,362 @@
...
+                if (tacoJson.kit) {
+                    delete tacoJson.kit;
+                }
+
+                if (!editParams.version) {
+                    deferred.reject(errorHelper.get(TacoErrorCodes.CommandCreateTacoJsonFileCreationError));
+                    return deferred.promise;
+                }
+
+                tacoJson["cordova-cli"] = editParams.version;
```
MSLaguana: Should this be cordovaCliVersion as well? What is the difference between editParams.version and cordovaCliVersion?

src/taco-cli/cli/tacoErrorCodes.ts
nisheetjain: good idea. DONE

src/taco-tests-utils/testProjectHelper.ts
nisheetjain: talked offline with @meenakb . Skipping this for now..

src/taco-utils/projectHelper.ts
nisheetjain: this requires some cleanup as part of another change. We really don't need tacoEditParams and we can just use tacoJsonMetadata here. version used to represent kit_version || cli_version but with recent changes (which always add cordova-cli field to taco.json) it would make more sense to rename the field. I will let @meenakb  drive that cleanup :)

PR #167, opened 2015-11-19
Return dsym from remotebuild

joshuaskelton:
## Summary
Enhancing remotebuild to return the symbol files for iOS device builds.

## Changes
* Refactored how we package build artifacts
 1. Zip symbol file directory
 2. Zip all artifacts (ipa, plist, dsym.zip)
 3. Send zip across the wire


src/taco-remote-lib/ios/ios.ts
```diff
@@ -113,37 +118,61 @@ class IOSAgent implements ITargetPlatform {
...
+            .then(function(): Q.Promise<any> {
+                var deferred: Q.Deferred<any> = Q.defer();
+                
+                var outputStream = fs.createWriteStream(pathToDsymZipFile);
+                var archive: any = archiver("zip");
+                archive.on("error", function (err: Error): void {
+                    return deferred.reject(err);
+                });
+                archive.on("finish", function() {
+                    return deferred.resolve({});
```
MSLaguana: I'm not sure, but there may be a slight race condition here: You are resolving the deferred when the archive stream finishes writing to the outputStream, but the outputStream may not yet have finished writing to disk. Might be best to resolve on outputStream's `end` event to guarantee that the next promise step can find the completed file on disk?

src/taco-remote-lib/package.json
```diff
@@ -28,7 +28,8 @@
         "rimraf": "2.2.6",
         "plist-with-patches": "0.5.1",
         "taco-utils": "file:../taco-utils",
-        "semver": "^4.3.6"
+        "semver": "^4.3.6",
+        "archiver": "^0.16.0"
```
MSLaguana: Do we still use zip-stream now? Or should that dependency be removed.

MSLaguana:
Looks good apart from the small chance of a race condition

src/taco-remote-lib/package.json
joshuaskelton: Good catch, zip-stream can probably be removed. I'll review it's usage.

PR #170, opened 2015-11-20
Now we set the error code name as the name of the TacoError, and we send it in the telemetry

digeff:
Error name reference: https://msdn.microsoft.com/en-us/library/96s4ewf9(v=vs.94).aspx


nisheetjain:
thanks for fixing this @digeff. Looks good to me

meenakb:
:+1:

PR #172, opened 2015-11-20
Refactoring flaky test suites to run reliably

meenakb:


src/taco-cli/test/kit.ts
```diff
@@ -214,67 +213,67 @@ describe("Kit command : ", function (): void {
...
-                telemetryParameters.should.be.eql(expected);
-                done();
-            }, function (err: tacoUtils.TacoError): void {
-                done(err);
-            });
-    });
+    describe("Kit list command : ", function (): void {
+        it("'taco kit' should not throw any error", function (done: MochaDone): void {
+            kitRun()
+                .done((telemetryParameters: TacoUtility.ICommandTelemetryProperties) => {
```
MSLaguana: Looks like we don't need the done handler, so we can change this to just return kitRun().then(...)

meenakb: Makes sense. @nisheetjain, can you review this?

src/taco-cli/test/kit.ts
```diff
@@ -214,67 +213,67 @@ describe("Kit command : ", function (): void {
...
         rimraf.sync(runFolder);
     });
 
-    after((done: MochaDone) => {
+    after() => {
         process.env["TACO_UNIT_TEST"] = previous;
         process.chdir(originalCwd);
         kitHelper.kitPackagePromise = null;
-        rimraf(runFolder, function (err: Error): void { done(); }); // ignore errors
+        rimraf.sync(runFolder); // ignore errors
```
MSLaguana: Is this the change that resolved the timeouts?

meenakb: Correct :)

PR #174, opened 2015-11-23
Livereload as a TACO package

omefire:
Moving Livereload into TACO as a package to streamline development and release efforts and make it easier.


src/taco-cli/cli/run.ts
```diff
@@ -88,6 +91,12 @@ class Run extends commands.TacoCommandBase {
             if (!remoteConfig) {
                 throw errorHelper.get(TacoErrorCodes.CommandRemotePlatformNotKnown, platform);
             }
+            
+            // DeviceSync/LiveReload not compatible with remote
+            var deviceSync = commandData.options["livereload"] || commandData.options["devicesync"];
+            if (deviceSync) {
```
meenakb: Why is this check needed here in addition to one in parseArgs?

omefire: Because, we can execute a remote platform in TACO without having to specify 'taco remote', It suffices to have a remote configuration set up. And when that's the case, the ```parseArgs``` code won't be sufficient to prevent us from running livereload on a remote target.

For example, on my Linux box, if I just issue: ```cordova run ios```, it runs it as remote platform.

axemclion: But what if the intention of the user is to run live reload on remote target too ? Will this prevent live-reload from running on remote targets ?

omefire: Currently, we are not supporting livereload on remote scenarios.
After discussing it, we realized that as the current remote architecture stands, it wouldn't work anyway and would require significant architectural changes to make it work. So, we left it out of this scope of work.

dblotsky:
Are the BrowserSync files new and up for review, or are they included as a library?

src/taco-livereload/lib/browserSyncServer.js
```diff
@@ -0,0 +1,134 @@
...
+ * @constructor
+ * @param {string} projectRoot - root folder of the project
+ * @param {Object} options - Options to initialize LiveReload with - These are to be passed to the underlying BrowserSync.
+ *                           See http://www.browsersync.io/docs/options/ for documentation
+ */
+
+function BrowserSyncServer(projectRoot, options) {
+    this.projectRoot = projectRoot;
+    this.options = options;
+    this.browserSync = require('browser-sync').create();
```
dblotsky: It's hard to find this dependency among the others. Might it be better to place it at the top with the others?

src/taco-cli/resources/en/resources.json
```diff
@@ -377,6 +377,12 @@
     "_CommandRunTargetDescription.comment": "Description of the --target argument to 'taco run'",
     "CommandRunListDescription": "Lists available targets. Will display both device and emulator targets unless overriden by --device or --emulator",
     "_CommandRunListDescription.comment": "Description of the --list argument to 'taco run'",
+    
+    "CommandRunLiveReloadDescription": "Run the application in Livereload mode",
```
meenakb: Please get these strings reviewed by @subhagpo :)

src/taco-livereload/lib/browserSyncServer.js
```diff
@@ -0,0 +1,134 @@
...
+                    return oldCanSync.apply(this, [data, optPath]);
+                };
+            }
+        } ());
+    };
+    return '<script>(' + script.toString() + '());</script>';
+}
+
+
+module.exports = BrowserSyncServer;
```
dblotsky: Please add a newline at the end of this file and any others that are missing newlines.

src/taco-livereload/package.json
```diff
@@ -0,0 +1,47 @@
...
+        "livereload",
+        "devicesync",
+        "browsersync"
+    ],
+    "main": "lib/index.js",
+    "preferGlobal": true,
+    "dependencies": {
+        "browser-sync": "2.7.13",
+        "q": "1.0.1",
+        "elementtree": "0.1.6",
```
nisheetjain: we mostly use ^a.b.c notation to allow revision updates

src/typings/tacoLiveReload.d.ts
```diff
@@ -0,0 +1,38 @@
...
+ *                                                     *
+ *******************************************************
+ */
+
+/// <reference path="../typings/Q.d.ts" />
+
+declare module TacoLiveReload {
+
+    export function start(projectRoot: string, platforms: string[], options: LiveReloadOptions): Q.Promise<any>;
+    export function on(eventName: string, cb: (...args : any[]) => void): void;
```
nisheetjain: what are the supported events here? should we put those in an enum?

src/typings/tacoLiveReload.d.ts
```diff
@@ -0,0 +1,38 @@
...
+    export function isLiveReloadActive(): boolean;
+
+    export class Patcher {
+        constructor(projectRoot: string, platforms: string[]);
+        public removeCSP(): void;
+        public patch(serverURL: string): Q.Promise<any>;
+    }
+
+    export interface LiveReloadHandle {
+        reloadFile(file?: string): Q.Promise<any>;
```
nisheetjain: I am not clear on what file argument does? what is the behavior in a multipage application? what if page is not even loaded in the app yet?

src/typings/tacoLiveReload.d.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/// <reference path="../typings/Q.d.ts" />
+
+declare module TacoLiveReload {
+
+    export function start(projectRoot: string, platforms: string[], options: LiveReloadOptions): Q.Promise<any>;
+    export function on(eventName: string, cb: (...args : any[]) => void): void;
+    export function isLiveReloadActive(): boolean;
+
+    export class Patcher {
```
nisheetjain: can you add some comments describing each class and methods

src/typings/tacoLiveReload.d.ts
omefire: Typescript currently doesn't support support string enums.
And doing something like the below example doesn't work. Any idea ?

```
    export class Events {
        static ERROR= "livereload:error";
    }
```

src/typings/tacoLiveReload.d.ts
omefire: - The 'file' argument is the file that we want to reload.
- In a multipage application, if possible, diffing of the file occurs and only changed parts will be reconstructed on the browser-side (e.g: .css). Otherwise, the full application gets reloaded (e.g: .js, .html, others).
- This handle will only be available after the app has loaded and started, so there's no risk of utilizing uninitialized resources.

src/taco-livereload/lib/browserSyncServer.js
omefire: Done! Thanks!

src/taco-livereload/lib/index.js
```diff
@@ -0,0 +1,22 @@
...
+ *                                                     *
+ *******************************************************
+ */
+
+/** This file is exposed externally */
+
+var livereload = require('./livereload');
+var helpers = require('./utils/helpers');
+
+module.exports = require('./livereload');
```
meenakb: Use the livereload var above

src/taco-livereload/package.json
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    "main": "lib/index.js",
+    "preferGlobal": true,
+    "dependencies": {
+        "browser-sync": "^2.7.13",
+        "q": "1.0.1",
+        "elementtree": "^0.1.6",
+        "underscore": "1.7.0",
+        "semver": "^4.3.x",
+        "shelljs": "^0.3.0",
```
meenakb: The above three packages do not seem to be used?

omefire:
@nisheetjain @meenakb If we are done reviewing this, can we merge it in ?

src/taco-cli/cli/run.ts
```diff
@@ -235,6 +244,18 @@ class Run extends commands.TacoCommandBase {
         if (parsedOptions.options["debug"] && parsedOptions.options["release"]) {
             throw errorHelper.get(TacoErrorCodes.ErrorIncompatibleOptions, "--debug", "--release");
         }
+        
+        if (parsedOptions.options["livereload"] && parsedOptions.options["remote"]) {
```
meenakb: Please watch out for merge conflicts with master (after Nisheet's refactoring commit)

src/taco-cli/cli/commands.json
```diff
@@ -444,6 +451,14 @@
             {
                 "name": "--target=TARGET", "description": "[CommandRunTargetDescription]"
             }
+            ,
+            {
+                 "name": "--livereload", "description": "[CommandRunLiveReloadDescription]"
```
meenakb: I see livereload/devicesync option handlers in run.ts only. You will have to make similar changes to emulate.ts as well.

omefire: Good catch. Fixed: https://github.com/Microsoft/TACO/commit/43957e176720b3b59ac579303d0d04469337819d

omefire:
I rebased this branch on top of the master branch and tested again.
We are good to go with the merge into master.

src/taco-cli/cli/run.ts
omefire: I rebased this branch on top of the master branch and tested again.
We are good to go with the merge into master.

src/taco-livereload/lib/livereload.js
```diff
@@ -0,0 +1,132 @@
...
+        var BrowserSyncServer = require('./browserSyncServer');
+        
+        // If user has entered whitespaces surrounding '--ignore', 
+        // ... we end up having helpers.parseOptions() return 'ignore' with value true instead of a string,
+        // ... which can't be used as a string. So, Let's check for those cases
+        // e.g of error cases: 
+        //      `cordova run android -- --livereload --ignore, // with no path to ignore
+        //      `cordova run android -- --livereload --ignore= css, // with space after the '=' sign
+        //      etc...
+        // ToDO: ignore whitespaces surrounding the --ignore option or at least display a warning message to user
```
meenakb: Please create appropriate tasks/bugs for these ToDo items and include them in comments.

src/taco-livereload/lib/browserSyncServer.js
```diff
@@ -0,0 +1,134 @@
...
+    var deferred = Q.defer();
+
+    var defaultOptions = {
+        server: {
+            baseDir: self.projectRoot,
+            directory: true
+        },
+        open: false,
+
+        // ToDO: Document that if fn is overriden by another option, multi-device scenario
```
meenakb: ToDo task/bug id needs to be included, for better tracking.

src/taco-livereload/lib/livereload.js
omefire: https://github.com/Microsoft/TACO/issues/190

meenakb:
:+1:

src/taco-utils/cordovaWrapper.ts
```diff
@@ -10,28 +10,34 @@
...
-import Q = require ("q");
-import semver = require ("semver");
-import util = require ("util");
+import assert = require("assert");
+import child_process = require("child_process");
+import path = require("path");
+import Q = require("q");
+import semver = require("semver");
+import util = require("util");
+import livereload = require("taco-livereload");
```
guillaumejenkins: You'll need to add `taco-livereload` to taco-utils package.json dependencies, otherwise this require will fail because it won't be able to find `taco-livereload`

omefire: Thanks @guillaumejenkins, added. The strange thing is that this works in certain environments with particular versions of node/npm.

guillaumejenkins: Yup, for those wondering: I suspect that with npm 3+, the new "flat" structure was causing taco-livereload to be installed at the root of node_modules (because taco-cli has a dependency on it), so taco-utils was able to require() it successfully even if it didn't have it as a dependency. However, for earlier nodes that still use the "nested" structure for dependency, taco-utils is not able to require() taco-cli's taco-livereload

PR #175, opened 2015-11-23
Changing taco-remote-lib to store the handle to idevicedebugserverpro…

MSLaguana:
…xy in a global variable to be explicitly shared between instances

idevicedebugserverproxy (more specifically the port that we configure it to use) is essentially a shared resource, and this change makes that explicit. If we end up running multiple instances of taco-remote-lib consecutively in order to handle multiple incompatible versions of Cordova, this change makes sure that the packages will check in a known, shared location to access the proxy. An identical change is needed on top of taco-remote-lib@1.2.0 to allow it to cooperate as well.

### DIFF
--- a/src/taco-remote-lib/ios/ios.ts
+++ b/src/taco-remote-lib/ios/ios.ts
@@ -33,7 +33,6 @@ import UtilHelper = utils.UtilHelper;
 
 class IOSAgent implements ITargetPlatform {
     private nativeDebugProxyPort: number;
-    private webProxyInstance: child_process.ChildProcess = null;
     private webDebugProxyDevicePort: number;
     private webDebugProxyPortMin: number;
     private webDebugProxyPortMax: number;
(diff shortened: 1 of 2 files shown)


MSLaguana:
See https://github.com/Microsoft/TACO/tree/fixConcurrentTacoRemoteLibPackages1.2.0 for the other side of things. Not sure how best to coordinate all this; don't really want that branch to merge into master, but we do want to keep it around so we can do more hotfixes / updates in future.

meenakb:
Looks good to me. Will you be sending another PR for hotfix on taco-remote-lib@1.2.0?

MSLaguana:
I'm not sure how best to do that; It shouldn't be merged into master / v1.2.0 branch, but it also doesn't make sense to be an addition to the v1.1.0 branch (since it has all the other fixes/changes made for the upcoming release as well, it just excludes some small set of things)

MSLaguana:
I am going to close this PR and open up another one against the v1.2.0 branch though.

PR #176, opened 2015-11-23
Updating development versions

joshuaskelton:


MSLaguana:
Looks like you missed taco-test-utils, but apart from that it looks good to me

meenakb:
:+1:

src/taco-remote/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "taco-remote",
     "description": "Module for the remotebuild package to build, run, and debug iOS apps created using Apache Cordova.",
-    "version": "1.2.1-dev",
+    "version": "1.3.1-dev",
```
guillaumejenkins: If we're going to bump the minor, should we reset the patch to 0?

src/taco-cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "taco-cli",
     "description": "taco-cli is a command-line interface for rapid Apache Cordova development",
-    "version": "1.0.3-dev",
+    "version": "1.1.1-dev",
```
guillaumejenkins: Same here, shouldn't the patch go back to 0? 1.1.**1**-dev

src/taco-utils/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "taco-utils",
     "description": "Common utilities used by the Tools for Apache Cordova node modules",
-    "version": "1.2.1-dev",
+    "version": "1.3.1-dev",
```
guillaumejenkins: Same

src/taco-remote/package.json
MSLaguana: This is correct, except the master branch here is one step ahead of when these versioning changes happened: the v1.2.0 branch has taco-remote version 1.3.0 in it. Master branch has "the next version" of taco-remote going forwards, so it has a larger version number, which will be corrected when we prepare the next build to publish (possibly remaining as a patch bump as this represents, possibly a minor or major bump if we make appropriate changes).

guillaumejenkins: Ah, right

PR #177, opened 2015-11-23
Fix concurrent taco remote lib packages

MSLaguana:
taco-remote-lib currently spawns two persistent processes for iOS related work: idevicedebugserverproxy and ios_webkit_debug_proxy. It keeps a hold of these processes in local variables, which starts to cause issues if we have multiple instances of the package around, such as now when we have different packages for cordova < 5.4.0 and >= 5.4.0

This change makes these processes into explicitly shared global state, allowing taco-remote-lib to share between packages.

This is one half of the change: identical changes are needed in taco-remote-lib@1.2.1

### DIFF
--- a/src/taco-remote-lib/ios/ios.ts
+++ b/src/taco-remote-lib/ios/ios.ts
@@ -24,6 +24,7 @@ import Packer = require ("zip-stream");
 
 import iosAppRunner = require ("./iosAppRunnerHelper");
 import resources = require ("../resources/resourceManager");
+import sharedState = require ("./sharedState");
 import utils = require ("taco-utils");
 
 import BuildInfo = utils.BuildInfo;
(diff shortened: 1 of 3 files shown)


MSLaguana:
See https://github.com/Microsoft/TACO/tree/fixConcurrentTacoRemoteLibPackages1.2.0 for the other package version. That is not intended to merge into this branch, it is more of a parallel branch.

src/taco-remote-lib/ios/iosAppRunnerHelper.ts
```diff
@@ -22,24 +22,26 @@ import utils = require ("taco-utils");
...
-var proxyInstance: child_process.ChildProcess = null;
 var promiseExec: (...args: any[]) => Q.Promise<any> = Q.denodeify(UtilHelper.loggedExec);
 
 class IosAppRunnerHelper {
     public static startDebugProxy(proxyPort: number): Q.Promise<child_process.ChildProcess> {
-        if (proxyInstance) {
-            proxyInstance.kill("SIGHUP"); // idevicedebugserver does not exit from SIGTERM
-            proxyInstance = null;
+        if (!global.tacoRemoteLib) {
+            global.tacoRemoteLib = {};
```
joshuaskelton: Why are we setting this? Is it not sufficient to add a guard to the statement below?

MSLaguana: This is lazily initializing the global.tacoRemoteLib object / namespace. If this is the first time we've ever tried to look at the tacoRemoteLib global storage, it won't exist, so we create it now. If we've run this method before (or another instance of this package has, e.g. taco-remote-lib@1.2.1) then the tacoRemoteLib global object will have been initialized already and we can proceed directly to using it.

src/taco-remote-lib/ios/ios.ts
```diff
@@ -230,14 +229,18 @@ class IOSAgent implements ITargetPlatform {
...
+
+        if (global.tacoRemoteLib.webProxyInstance) {
+            global.tacoRemoteLib.webProxyInstance.kill();
+            global.tacoRemoteLib.webProxyInstance = null;
         }
 
         var portRange: string = util.format("null:%d,:%d-%d", this.webDebugProxyDevicePort, this.webDebugProxyPortMin, this.webDebugProxyPortMax);
         try {
-            this.webProxyInstance = child_process.spawn("ios_webkit_debug_proxy", ["-c", portRange]);
+            global.tacoRemoteLib.webProxyInstance = child_process.spawn("ios_webkit_debug_proxy", ["-c", portRange]);
```
dlebu: Do we need an `.on("error"` handler for this one? I see you have a handler for `proxyInstance`.

MSLaguana: If it lacks an `on("error"` handler then it will throw an exception, but we have a try/catch here which I believe should handle it. I think it's something worth doing, but I'm not sure whether it is worth mixing in with the ask-mode fix?

src/taco-remote-lib/ios/ios.ts
```diff
@@ -230,14 +229,18 @@ class IOSAgent implements ITargetPlatform {
     }
 
     public debugBuild(buildInfo: utils.BuildInfo, req: Express.Request, res: Express.Response): void {
-        if (this.webProxyInstance) {
-            this.webProxyInstance.kill();
-            this.webProxyInstance = null;
+        if (!global.tacoRemoteLib) {
```
dlebu: I feel it would be nice to have a single place where we manage all the globals. Does it make sense to extract the globals management code to a different class?

src/taco-remote-lib/ios/ios.ts
dlebu: If you are sure about the exception, then this is fine.

dlebu:
Minor comments, LGTM otherwise.

MSLaguana:
Note that prior to release, we'll have to modify the taco-remote-multiplexer dynamicdependencies.json to point to the yet-to-be-released taco-remote-lib@1.2.1

PR #179, opened 2015-11-24
Fixed errors not being localized if they don't have args

guillaumejenkins:


src/taco-utils/tacoError.ts
```diff
@@ -76,7 +76,7 @@ module TacoUtility {
                     message = resources.getString(errorToken, args);
                 }
             } else {
```
MSLaguana: Do you know if this conditional is necessary? this function was just `return resources.getString(errorToken, args)` does anything misbehave? I thought that the ResourcesManager could deal with an empty array, but I wasn't sure.

guillaumejenkins: I thought about that at first, but then I saw that `args` is an array, not a `...args: any[]`. So, I'm afraid that if I pass `args` directly to `getString()`, it will treat the array as the first argument rather than as a list of several arguments. If that's the case, it means we would need to do some massaging to invoke `getString()` with each element of the array as its own separate argument.

If there is an easy way to just pass an array to a `...args: any[]` type, then we could remove this method entirely and use `resources.getString()` directly in the callers

guillaumejenkins: It looks like I can just remove this method entirely and use the "..." notation to invoke `resources.getString()` directly

MSLaguana:
Looks good to me.

joshuaskelton:
Looks good.

dlebu:
LGTM!

PR #185, opened 2015-11-25
parseUserJSON call got removed as part of refactoring. Adding it back.

nisheetjain:
Also added a test case to handle this case


meenakb:
Looks good to me.

PR #186, opened 2015-11-30
Remove kits from create tests

guillaumejenkins:
We were explicitly requiring taco-kits in the create tests; I changed this to be lazily required

### DIFF
--- a/src/taco-cli/test/create.ts
+++ b/src/taco-cli/test/create.ts
@@ -46,7 +46,6 @@ import TemplateManager = require ("../cli/utils/templateManager");
 import tacoTestsUtils = require ("taco-tests-utils");
 
 import IKeyValuePair = tacoTestUtils.IKeyValuePair;
-import TacoKitsErrorCodes = TacoKits.TacoErrorCode;
 import TacoUtilsErrorCodes = tacoUtils.TacoErrorCode;
 import utils = tacoUtils.UtilHelper;
 import MemoryStream = tacoTestsUtils.MemoryStream;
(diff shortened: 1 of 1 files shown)


src/taco-cli/test/create.ts
```diff
@@ -227,14 +227,15 @@ describe("taco create", function (): void {
...
-            }
+        rimraf.sync(runFolder);
+
+        // Create the run folder for our tests
+        wrench.mkdirSyncRecursive(runFolder, 511); // 511 decimal is 0777 octal
+
+        // Load taco-kits error codes
+        tacoPackageLoader.lazyRequire("taco-kits", "taco-kits").then((tacoKits: typeof TacoKits) => {
+            tacoKitsErrors = tacoKits.TacoErrorCode;
+            done();
```
MSLaguana: I believe that you can not take `done` as an argument, and just return this promise here. You could even promise chain the rimraf and wrench together with the lazy require (or do them all concurrently)

src/taco-cli/test/create.ts
```diff
@@ -47,7 +46,7 @@ import TemplateManager = require ("../cli/utils/templateManager");
 import tacoTestsUtils = require ("taco-tests-utils");
 
 import IKeyValuePair = tacoTestUtils.IKeyValuePair;
-import TacoKitsErrorCodes = tacoKits.TacoErrorCode;
```
MSLaguana: How did this work before when taco-cli did not take any dependency on taco-kits in package.json?

Oh, it's because our folder structure had things in a root node_modules folder, so it just happened to work.

MSLaguana:
It feels a little strange to do a lazy require of taco-kits just to get a couple of error codes out of it, but they'll be lazy required by the code under test anyway so it will be cached anyway. Looks good to me.

meenakb:
:+1:

PR #187, opened 2015-11-30
Improving ICommand and ISubCommand interfaces

nisheetjain:
TacoCommandBase can now do a simple name based comparison to route
subcommands
ISubCommand has access to parent ICommand for alias resolving and
redirection
ICommand.run takes string args

### DIFF
--- a/src/taco-cli/cli/build.ts
+++ b/src/taco-cli/cli/build.ts
@@ -137,7 +137,7 @@ class Build extends commands.TacoCommandBase {
         });
     }
 
-    private static build(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
+     protected runCommand(commandData: commands.ICommandData): Q.Promise<ICommandTelemetryProperties> {
         if (ProjectHelper.isTypeScriptProject()) {
             logger.log(resources.getString("CommandCreateInstallGulp"));
         }
@@ -161,19 +161,6 @@ class Build extends commands.TacoCommandBase {
             });
         }).then(() => Build.generateTelemetryProperties(telemetryProperties, commandData));
     }
-
-    /* tslint:disable:member-ordering */
-    // tslint doesn't handle this case and considers subcommands as member function
-    public subcommands: commands.ICommand[] = [
-        {
-            name: "build",
-            run: Build.build,
-            canHandleArgs(commandData: commands.ICommandData): boolean {
-                return true;
-            }
-        }
-    ];
-    /* tslint:enable:member-ordering */
 
     /**
      * specific handling for whether this command can handle the args given, otherwise falls through to Cordova CLI
--- a/src/taco-cli/cli/build.ts
+++ b/src/taco-cli/cli/build.ts
@@ -137,7 +137,7 @@ class Build extends commands.TacoCommandBase {
         });
     }
 
-     protected runCommand(commandData: commands.ICommandData): Q.Promise<ICommandTelemetryProperties> {
+    protected runCommand(commandData: commands.ICommandData): Q.Promise<ICommandTelemetryProperties> {
         if (ProjectHelper.isTypeScriptProject()) {
             logger.log(resources.getString("CommandCreateInstallGulp"));
         }
--- a/src/taco-cli/cli/kit.ts
+++ b/src/taco-cli/cli/kit.ts
@@ -779,22 +779,19 @@ class Kit extends commands.TacoCommandBase {
 
     /* tslint:disable:member-ordering */
     // tslint doesn't handle this case and considers subcommands as member function
-    public subcommands: commands.ISubCommand<Kit>[] = [
-        {
-            // List kits
-            name: "list",
-            run: (command, commandData) => command.list(commandData),
-            canHandleArgs: (command, commandData) => {
-                var subCommand: string = commandData.remain[0];
-                return !subCommand || command.resolveAlias(subCommand) === this.name;
-            }
-        },
+    public subcommands: commands.ISubCommand[] = [
         {
             // Change kit or CLI
             name: "select",
-            run: (command, commandData) => command.select(commandData)
+            run: commandData => this.select(commandData)
         },
-    ];
+        {
+            // List kits
+            name: "list",
+            run: commandData => this.list(commandData),
+            canHandleArgs: commandData => true
+        }
+   ];
     /* tslint:enable:member-ordering */
 
     public parseArgs(args: string[]): commands.ICommandData {
--- a/src/taco-cli/cli/kit.ts
+++ b/src/taco-cli/cli/kit.ts
@@ -721,7 +721,7 @@ class Kit extends commands.TacoCommandBase {
         });
     }
 
-    private static select(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
+    private select(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
         var kitId: string = commandData.options["kit"];
         var cli: string = commandData.options["cordova"];
         var projectInfo: IProjectInfo;
@@ -760,7 +760,7 @@ class Kit extends commands.TacoCommandBase {
         }).then(() => Kit.generateTelemetryProperties(commandData));
     }
 
-    private static list(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
+    private list(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
         logger.logLine();
         var kitId: string = commandData.options["kit"];
         var jsonPath: any = commandData.options["json"];
@@ -779,22 +779,20 @@ class Kit extends commands.TacoCommandBase {
 
     /* tslint:disable:member-ordering */
     // tslint doesn't handle this case and considers subcommands as member function
-    public subcommands: commands.ICommand[] = [
+    public subcommands: commands.ISubCommand<Kit>[] = [
         {
             // List kits
             name: "list",
-            run: Kit.list,
-            canHandleArgs(commandData: commands.ICommandData): boolean {
-                return !commandData.remain[0] || commandData.remain[0] && commandData.remain[0].toLowerCase() === "list";
+            run: (command, commandData) => command.list(commandData),
+            canHandleArgs: (command, commandData) => {
+                var subCommand: string = commandData.remain[0];
+                return !subCommand || command.resolveAlias(subCommand) === this.name;
             }
         },
         {
             // Change kit or CLI
             name: "select",
-            run: Kit.select,
-            canHandleArgs(commandData: commands.ICommandData): boolean {
-                return !commandData.remain[0] || commandData.remain[0] && commandData.remain[0].toLowerCase() === "select";
-            }
+            run: (command, commandData) => command.select(commandData)
         },
     ];
     /* tslint:enable:member-ordering */
--- a/src/taco-cli/cli/kit.ts
+++ b/src/taco-cli/cli/kit.ts
@@ -777,8 +777,6 @@ class Kit extends commands.TacoCommandBase {
         return result.then(() => Kit.generateTelemetryProperties(commandData));
     }
 
-    /* tslint:disable:member-ordering */
-    // tslint doesn't handle this case and considers subcommands as member function
     public subcommands: commands.ISubCommand[] = [
         {
             // Change kit or CLI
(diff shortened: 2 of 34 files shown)


nisheetjain:
key changes are in Commands.ts.

src/taco-cli/cli/build.ts
```diff
@@ -137,7 +137,7 @@ class Build extends commands.TacoCommandBase {
         });
     }
 
-    private static build(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
+     protected runCommand(commandData: commands.ICommandData): Q.Promise<ICommandTelemetryProperties> {
```
MSLaguana: rogue space here?

src/taco-cli/cli/kit.ts
```diff
@@ -721,7 +721,7 @@ class Kit extends commands.TacoCommandBase {
         });
     }
 
-    private static select(commandData: commands.ICommandData): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
```
MSLaguana: What's the reasoning behind making these instance methods rather than static methods?

src/taco-cli/cli/kit.ts
```diff
@@ -779,22 +779,20 @@ class Kit extends commands.TacoCommandBase {
...
-    public subcommands: commands.ICommand[] = [
+    public subcommands: commands.ISubCommand<Kit>[] = [
         {
             // List kits
             name: "list",
-            run: Kit.list,
-            canHandleArgs(commandData: commands.ICommandData): boolean {
-                return !commandData.remain[0] || commandData.remain[0] && commandData.remain[0].toLowerCase() === "list";
+            run: (command, commandData) => command.list(commandData),
+            canHandleArgs: (command, commandData) => {
```
MSLaguana: I was a little confused for a moment about what `this` would be since the fat arrow here isn't in a context where `this` already exists. It ends up being fine, but is there a reason you preferred it over the other style?

src/taco-cli/cli/kit.ts
nisheetjain: because we should avoid full static implementations. they can't be extended. for e.g. for remote say I want to add "taco macincloud add". I would want to extend existing implementation.

MSLaguana: Ok, fair point.

src/taco-cli/cli/kit.ts
nisheetjain: this would point to the subcommand and that is the reason we need to pass along command object

MSLaguana: Are you sure about that? When I tested this out the `this` was referring to the class that declared the function, which is `Kit` here.

```
class A {
    public static name: string = "A";
    public name: string = "a";
    public thing = () => {
        console.info(this.name);
    }
    public things = [
        {
            name: "c",
            f: () => console.info(this.name)
        }]
}

var a = new A();
a.thing();
a.things[0].f();
```

prints out `aa` meaning that `this` referred to the instance of the class both times. If the fat arrow function in the `things` list was instead `f(){ console.info(this.name)}` then `this` referred to that object (because the function was invoked from that object).

Now I suspect that the code is currently incorrect with the usage of `this`, and it will refer to `Kit.name` and not the subcommand's name.

src/taco-cli/cli/remote.ts
```diff
@@ -59,38 +59,27 @@ class Remote extends commands.TacoCommandBase {
...
         },
         {
             // taco remote [unknown]
             name: "help",
-            run: Remote.help,
-            canHandleArgs: function (remoteData: commands.ICommandData): boolean {
-                return true;
-            }
+            run: (command, commandData) => command.help(commandData),
+            canHandleArgs: (command, commanData) => true
```
MSLaguana: Typo `commanData`

src/taco-cli/cli/taco.ts
```diff
@@ -73,16 +73,13 @@ class Taco {
...
 
             var parsedArgs: IParsedArgs = Taco.parseArgs(process.argv.slice(2));
-            var commandProperties: ICommandTelemetryProperties = {};
 
             Taco.runWithParsedArgs(parsedArgs)
                 .then(function (telemetryProperties: ICommandTelemetryProperties): void {
                     if (parsedArgs.command) {
-                        commandProperties = telemetryProperties;
+                        // Send command success telemetry
+                        telemetryHelper.sendCommandSuccessTelemetry(parsedArgs.commandName, telemetryProperties, parsedArgs.args);
```
MSLaguana: With this change we will no longer send success telemetry if `parsedArgs.command` is falsy, is that deliberate?

src/taco-cli/test/buildAndRunTelemetry.ts
```diff
@@ -291,7 +291,12 @@ module BuildAndRunTelemetryTests {
             (typeof telemetryProperties === "undefined").should.be.equal(false);
             validateGzipedSize(telemetryProperties, "ios", iosExpectedGzipedSize);
             validateGzipedSize(telemetryProperties, "android", androidGzipSize);
-            telemetryProperties.should.eql(expected); // We are comparing the objects, after overriding the sizes with the expected values
+
+            // validated that each of the expected telemetry properties are present
+            for (var key in expected) {
+                console.log(key);
```
MSLaguana: I assume this log should be removed

src/taco-cli/test/buildAndRunTelemetry.ts
```diff
@@ -291,7 +291,12 @@ module BuildAndRunTelemetryTests {
             (typeof telemetryProperties === "undefined").should.be.equal(false);
             validateGzipedSize(telemetryProperties, "ios", iosExpectedGzipedSize);
             validateGzipedSize(telemetryProperties, "android", androidGzipSize);
-            telemetryProperties.should.eql(expected); // We are comparing the objects, after overriding the sizes with the expected values
```
MSLaguana: Why did you change from the `.eql` format? Do you want to allow telemetryProperties to have keys that are not in expected without failing the test?

src/taco-cli/cli/build.ts
meenakb: How about converting the other static methods in this class to instance methods as well - buildRemotePlatform, cleanPlatform?

src/taco-utils/commands.ts
```diff
@@ -50,14 +50,19 @@ module TacoUtility {
          */
         export interface ICommand {
             name: String;
-            run(data: ICommandData): Q.Promise<ICommandTelemetryProperties>;
+            run(args: string[]): Q.Promise<ICommandTelemetryProperties>;
             canHandleArgs(data: ICommandData): boolean;
         }
 
+        export interface ISubCommand<T extends ICommand> {
```
MSLaguana: Did you intend to replace instances of `ICommand` within this interface with the specific `T`?

src/taco-cli/cli/build.ts
nisheetjain: yes. I would love to have that. But I would let the code owner drive that cleanup..

src/taco-cli/cli/remote.ts
nisheetjain: done

src/taco-utils/commands.ts
```diff
@@ -77,36 +82,49 @@ module TacoUtility {
...
 
             public resolveAlias(subCommand: string): string {
                 return (this.info.aliases && this.info.aliases[subCommand]) ? this.info.aliases[subCommand] : subCommand;
             }
 
-            private getSubCommand(options: ICommandData): ICommand {
+            private getSubCommand(options: ICommandData): ISubCommand<ICommand> {
+                // first do a simple name match
+                for (var i: number = 0; i < this.subcommands.length; ++i) {
+                    var name: string = options.remain[0];
```
MSLaguana: You can move the name outside of the for loop

src/taco-cli/cli/kit.ts
nisheetjain: you are right. Removing the command param in next commit

src/taco-cli/cli/remote.ts
```diff
@@ -59,38 +59,27 @@ class Remote extends commands.TacoCommandBase {
...
-    public subcommands: commands.ICommand[] = [
+    public subcommands: commands.ISubCommand<Remote>[] = [
         {
             // taco remote remove <platform>
             name: "remove",
-            run: Remote.remove,
-            canHandleArgs: function (remoteData: commands.ICommandData): boolean {
-                return remoteData.remain[0] && /^(remove|rm)$/i.test(remoteData.remain[0]);
-            }
+            run: (command, commandData) => command.remove(commandData)
```
meenakb: Missing overrides for CanHandleArgs.

src/taco-cli/cli/kit.ts
```diff
@@ -779,22 +779,20 @@ class Kit extends commands.TacoCommandBase {
...
             }
         },
         {
             // Change kit or CLI
             name: "select",
-            run: Kit.select,
-            canHandleArgs(commandData: commands.ICommandData): boolean {
-                return !commandData.remain[0] || commandData.remain[0] && commandData.remain[0].toLowerCase() === "select";
-            }
+            run: (command, commandData) => command.select(commandData)
```
meenakb: Here as well (override for CanHandleArgs)

src/taco-cli/test/buildAndRunTelemetry.ts
nisheetjain: yes because test is written in a way where it executes for build/run/emulate. Run I believe has subcommand in it, rest two don't have it.

src/taco-cli/test/buildAndRunTelemetry.ts
nisheetjain: good catch :)

src/taco-utils/commands.ts
nisheetjain: removed generic part out of subcommand

src/taco-cli/test/buildAndRunTelemetry.ts
MSLaguana: Might be able to use `telemetryProperties.should.containEql(expected)`? Looks like `{x: 1, y: 2}.should.containEql({y:2})` works and `{x: 1, y: 2}.should.containEql({z:3})` fails, so I think that's what we want here

src/taco-cli/cli/taco.ts
meenakb: We already handle that case (routed commands to Cordova) in runWithParsedArgs. So, this change looks good.

src/taco-cli/cli/remote.ts
nisheetjain: discussed offline. not needed since we are mapping remove directly

src/taco-cli/cli/run.ts
```diff
@@ -177,38 +177,30 @@ class Run extends commands.TacoCommandBase {
 
     /* tslint:disable:member-ordering */
```
guillaumejenkins: How that the other methods are not longer static, can this rule be deactivated and this member be put at the top of the file?

meenakb:
:+1: cool stuff

src/taco-cli/cli/run.ts
nisheetjain: makes sense :)

MSLaguana:
Looks good to me.

src/taco-utils/commands.ts
```diff
@@ -50,14 +50,19 @@ module TacoUtility {
          */
         export interface ICommand {
             name: String;
-            run(data: ICommandData): Q.Promise<ICommandTelemetryProperties>;
+            run(args: string[]): Q.Promise<ICommandTelemetryProperties>;
             canHandleArgs(data: ICommandData): boolean;
         }
 
+        export interface ISubCommand {
```
guillaumejenkins: Should this interface extend `ICommand`? Slightly cleaner I think, that way you don't have to duplicate `name` and `canHandleArgs`, and it should still let you override `run`

src/taco-utils/commands.ts
```diff
@@ -77,36 +82,50 @@ module TacoUtility {
...
+                name = name ? this.resolveAlias(name) : name;
                 for (var i: number = 0; i < this.subcommands.length; ++i) {
-                    var subCommand: ICommand = this.subcommands[i];
-                    if (subCommand.canHandleArgs(options)) {
+                    if (name === this.subcommands[i].name) {
+                        return this.subcommands[i];
+                    }
+                }
+                // if subcommand has specified a canHandleArgs, check if there is a subcommand
+                // which can handle the args
```
guillaumejenkins: I'm not sure I understand the intent behind this logic... Why would we try to run a subcommand if the name doesn't match? Does this mean that if `taco kit select` implements `canHandleArgs()` and we do `taco kit dasdasdsa --cordova 5.2.0`, we will end up running `select`? If that's the case, then I feel it might be a bit dangerous to be so permissive

PR #188, opened 2015-11-30
--accept parameter

guillaumejenkins:
Added a global ~~--accept~~ **Edit: changed to --no-prompt** parameter that gets saved in TacoGlobalConfig. This can be used by packages to react accordingly when prompting the user.

Also made DependencyInstaller honor that parameter.

### DIFF
--- a/src/taco-dependency-installer/dependencyInstaller.ts
+++ b/src/taco-dependency-installer/dependencyInstaller.ts
@@ -375,7 +375,7 @@ module TacoDependencyInstaller {
             }
 
             // If we accept prompts automatically, then return immediately
-            if (TacoGlobalConfig.acceptPrompts) {
+            if (TacoGlobalConfig.noPrompt) {
                 return Q.resolve(true);
             }
 
@@ -518,7 +518,7 @@ module TacoDependencyInstaller {
                         utilHelper.quotesAroundIfNecessary(self.installConfigFilePath),
                         self.parentSessionId,
                         utilHelper.quotesAroundIfNecessary(DependencyInstaller.socketPath),
-                        TacoGlobalConfig.acceptPrompts ? "true" : "false"   // When launching a child process inside PowerShell, the environment is not preserved, so we need to pass this value on the command line
+                        TacoGlobalConfig.noPrompt ? "true" : "false"   // When launching a child process inside PowerShell, the environment is not preserved, so we need to pass this value on the command line
                     ];
                     var cp: childProcess.ChildProcess = childProcess.spawn(command, args, { stdio: "ignore" }); // Note: To workaround a Powershell hang on Windows 7, we set the stdio to ignore, otherwise Powershell never returns
 
--- a/src/taco-dependency-installer/dependencyInstaller.ts
+++ b/src/taco-dependency-installer/dependencyInstaller.ts
@@ -518,7 +518,7 @@ module TacoDependencyInstaller {
                         utilHelper.quotesAroundIfNecessary(self.installConfigFilePath),
                         self.parentSessionId,
                         utilHelper.quotesAroundIfNecessary(DependencyInstaller.socketPath),
-                        TacoGlobalConfig.acceptPrompts ? "true" : "false"
+                        TacoGlobalConfig.acceptPrompts ? "true" : "false"   // When launching a child process inside PowerShell, the environment is not preserved, so we need to pass this value on the command line
                     ];
                     var cp: childProcess.ChildProcess = childProcess.spawn(command, args, { stdio: "ignore" }); // Note: To workaround a Powershell hang on Windows 7, we set the stdio to ignore, otherwise Powershell never returns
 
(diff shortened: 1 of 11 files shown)


src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -109,6 +110,10 @@ module TacoDependencyInstaller {
                 telemetry.step("promptUserBeforeInstall");
                 return this.promptUserBeforeInstall()
                     .then(function (acceptedPrompt: boolean): Q.Promise<number> {
+                        logger.logLine();
+                        loggerHelper.logSeparatorLine();
```
MSLaguana: Why is the spacing changing?

src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -369,15 +374,16 @@ module TacoDependencyInstaller {
                 logger.log(resources.getString("LicenseAgreement"));
             }
 
+            // If we accept prompts automatically, then return immediately
+            if (TacoGlobalConfig.acceptPrompts) {
+                return Q.resolve(true);
```
MSLaguana: I'm inclined to print out a line stating that the prompt was automatically accepted, rather than skipping it entirely.

src/taco-dependency-installer/dependencyInstaller.ts
```diff
@@ -511,7 +517,8 @@ module TacoDependencyInstaller {
                         utilHelper.quotesAroundIfNecessary(elevatedInstallerPath),
                         utilHelper.quotesAroundIfNecessary(self.installConfigFilePath),
                         self.parentSessionId,
-                        utilHelper.quotesAroundIfNecessary(DependencyInstaller.socketPath)
+                        utilHelper.quotesAroundIfNecessary(DependencyInstaller.socketPath),
+                        TacoGlobalConfig.acceptPrompts ? "true" : "false"
```
MSLaguana: Might want to add a note that we pass the value along on the command line even though it is set in an environment variable because environment variables are not maintained through process elevation.

src/taco-dependency-installer/utils/win32/elevatedInstallerLauncher.ps1
```diff
@@ -4,7 +4,7 @@ $pinfo.FileName = "node.exe"
 $pinfo.UseShellExecute = $true
 $pinfo.WindowStyle = "Hidden"
 $pinfo.Verb = "RunAs"
-$pinfo.Arguments = "$($args[0]) $($args[1]) $($args[2]) $($args[3])"
+$pinfo.Arguments = "$args
```
MSLaguana: Looks like this is broken; I assume it should be passing through another argument or so rather than removing a bunch?

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: This separator used to be logged after we returned from prompting the user; it had the consequence of not being printed when we didn't prompt, so I moved it to a common code path

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: Interesting, I'll take this to Soak

src/taco-dependency-installer/utils/win32/elevatedInstallerLauncher.ps1
guillaumejenkins: I've just learned that in PowerShell, if you put an array inside quotes, it automatically does a join of the array values into a space-separated string. I've changed it to this way so that we don't have to modify the PowerShell script every time we change what is exchanged between the elevated installer and the dependency installer

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: So, decision is to not print such messages (as user should be expecting that and went out of his way to skip prompts), but I should have added a help entry to explain this new parameter, so I'll do that now

src/taco-dependency-installer/dependencyInstaller.ts
guillaumejenkins: Will do

nisheetjain:
what happened to supporting "echo y|taco install-reqs" instead?

src/taco-dependency-installer/elevatedInstaller.ts
```diff
@@ -103,6 +104,12 @@ class ElevatedInstaller {
         this.configFile = process.argv[2];
         this.parentSessionId = process.argv[3];
         this.socketPath = process.argv[4];
+
+        // Deal with auto-accept prompts; we need to reset this value, because on Windows we launch the elevated installer through PowerShell, and doing it that way does not carry over the env of the
+        // parent node.js process
+        if (process.argv[5] === "true") {
```
nisheetjain: why don't we do TacoGlobalConfig.noPrompt here? wouldn't elevatedInstaller inherit the env block?

nisheetjain:
LGTM otherwise

src/taco-dependency-installer/elevatedInstaller.ts
MSLaguana: The elevation process strips out most environment variables to avoid security issues.

guillaumejenkins: What Jimmy said; On Mac we can use "sudo -E" to keep the environment through elevation, but it doesn't look like there is something similar for Windows. So, we pass the noPrompt flag as a command line parameter, and then we set tacoGlobalConfig.noPrompt again in the elevated installer based on that

nisheetjain: makes sense. LG :)

PR #195, opened 2015-12-03
Changing definition of 'eligible for deletion' to only exclude builds in progress

MSLaguana:
Currently the criteria for deleting a build excludes "installed" or "debugging" builds, which mean they clog up the pipeline. The more specific criteria below excludes only builds that are currently queued for building or actually building.

### DIFF
--- a/src/taco-remote/lib/buildRetention.ts
+++ b/src/taco-remote/lib/buildRetention.ts
@@ -60,9 +60,7 @@ class BuildRetention {
         var eligibleBuilds: utils.BuildInfo[] = [];
         for (var i: number = 0; i < buildNumbers.length; i++) {
             var buildInfo: utils.BuildInfo = builds[buildNumbers[i]];
-            if (buildInfo.status === utils.BuildInfo.COMPLETE || buildInfo.status === utils.BuildInfo.DOWNLOADED ||
-                buildInfo.status === utils.BuildInfo.ERROR || buildInfo.status === utils.BuildInfo.EMULATED ||
-                buildInfo.status === utils.BuildInfo.INVALID) {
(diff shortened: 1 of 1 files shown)


src/taco-remote/lib/buildRetention.ts
```diff
@@ -60,9 +60,7 @@ class BuildRetention {
         var eligibleBuilds: utils.BuildInfo[] = [];
         for (var i: number = 0; i < buildNumbers.length; i++) {
             var buildInfo: utils.BuildInfo = builds[buildNumbers[i]];
-            if (buildInfo.status === utils.BuildInfo.COMPLETE || buildInfo.status === utils.BuildInfo.DOWNLOADED ||
-                buildInfo.status === utils.BuildInfo.ERROR || buildInfo.status === utils.BuildInfo.EMULATED ||
-                buildInfo.status === utils.BuildInfo.INVALID) {
+            if ([utils.BuildInfo.BUILDING, utils.BuildInfo.EXTRACTED, utils.BuildInfo.UPLOADING, utils.BuildInfo.UPLOADED].indexOf(buildInfo.status) !== -1) {
```
nisheetjain: can we put this array as a static property with meaningful name? in fact if these were bits we could just done an "and"? or may be define BuildInfo.InProgress

dlebu: Another way we could solve this is to implement `BuildInfo` a Java style enum. In Java, enums can have methods as well, so you can save state in each enum entry. For this case, we could have something like `buildInfo.status.pushEligible()`.

dlebu:
LGTM otherwise!

PR #200, opened 2015-12-08
Adding some promises syntactic sugar methods like chain, condition, and,

nisheetjain:
or


src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,101 @@ module TacoUtility {
...
+         *  @value: array of values used to create chain of promises
+         *  @func: accumulator function which runs over each array value and returns a promise which resolves to an accumulated value
+         *  @initialValue: initial accumulated value
+         */
+        public static chain<T, U>(values: T[], func: (value: T, valueSoFar: U) => Q.Promise<U>, initialValue?: U): Q.Promise<U> {
+            return values.reduce(function(soFar: Q.Promise<U>, val: T): Q.Promise<U> {
+                return soFar.then(function(valueSoFar: U): Q.Promise<U> {
+                    return func(val, valueSoFar);
+                });
+            }, Q(initialValue || <U>{}));
```
MSLaguana: Why not just `Q(initialValue)`? If U is supposed to be a numeric type for example, then {} is incorrect. If you just leave it as initialValue, then it will go to undefined (which inhabits all types and the code should be willing to deal with).

src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,101 @@ module TacoUtility {
...
+        }
+
+        /**
+         * Syntactic sugar for promise and
+         *  @conditions: a variable array of promises which resolve to true/false
+         *
+         *  @returns: resolves to true if both the promises resolve to true, false otherwise
+         */
+        public static and(...conditions: PromiseFuncOrValue<boolean>[]): Q.Promise<boolean> {
+            return PromisesUtils.logicalOp(true, ...conditions);
```
MSLaguana: Should be false rather than true, I believe?

nisheetjain: good catch. copy paste error, and should have false

src/taco-cli/test/checkForNewerVersion.ts
```diff
@@ -53,8 +53,6 @@ describe("Check for newer version", function (): void {
 
     var expectedRequestAndResponse: { expectedUrl: string; statusCode: number; head: any; response: any; waitForPayload?: boolean, responseDelay?: number };
 
-    // We should be able to remove the next line, after this fix gets released: https://github.com/mochajs/mocha/issues/779
-    this.timeout(15000);
```
MSLaguana: Why are these happening here?

nisheetjain: I did "git rebase master"

src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,95 @@ module TacoUtility {
...
+            promiseTrue: (() => Q.Promise<T>) | T,
+            promiseFalse: (() => Q.Promise<T>) | T): Q.Promise<T> {
+
+            var _promiseTrue: () => Q.Promise<T>  = PromisesUtils.getPromiseFunc(promiseTrue);
+            var _promiseFalse: () => Q.Promise<T> = PromisesUtils.getPromiseFunc(promiseFalse);
+
+            if (typeof condition === "boolean") {
+                return condition ? _promiseTrue() : _promiseFalse();
+            }
+            return (<Q.Promise<boolean>>condition)
```
MSLaguana: I don't think this cast is required, since Typescript understands type guards like the one you have above.

MSLaguana: Alternately, you could just always wrap it, collapsing both cases together with something like `return Q(condition).then(function(result: Boolean): ...`

src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,95 @@ module TacoUtility {
...
+        }
+
+        /**
+         * Syntactic sugar for if/else style promises
+         *  @condition: Promise resolving to true/false 
+         *  @promiseTrue: then promise if condition resolves to "true"
+         *  @promiseFalse: then promise if condition resolves to "false"
+         */
+        public static condition<T>(condition: Q.Promise<boolean> | boolean,
+            promiseTrue: (() => Q.Promise<T>) | T,
```
MSLaguana: These types could also be `PromiseFuncOrVal<T>`, right?

src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,95 @@ module TacoUtility {
...
+
+        /**
+         * Syntactic sugar for promise or
+         *  @condition1: Promise1 resolving to true/false 
+         *  @condition2: Promise2 resolving to true/false 
+         *
+         *  @returns: resolves to true if either of the promises resolve to true, false otherwise
+         */
+        private static logicalOp(exitValue: boolean, ...conditions: PromiseFuncOrValue<boolean>[]): Q.Promise<boolean> {
+            var args: PromiseFuncOrValue<boolean>[] = <PromiseFuncOrValue<boolean>[]>ArgsHelper.getOptionalArgsArrayFromFunctionCall(arguments, 1);
```
MSLaguana: I think we should try to avoid using getOptionalArgsArrayFromFunctionCall if we can help it. Unless you want to explicitly support `logicalOp([p1, p2, p3])` and `logicalOp(p1, p2, p3)`, we should just pick one.

src/taco-utils/promisesUtils.ts
```diff
@@ -26,12 +30,95 @@ module TacoUtility {
...
+            return args.reduce(function(soFar: Q.Promise<boolean>, arg: () => Q.Promise<boolean>): Q.Promise<boolean> {
+                return soFar.then(function(valueSoFar: boolean): Q.Promise<boolean> {
```
MSLaguana: If we want complete short circuiting behaviour without stepping through the remainder of the list, we could do something like

```
var stepFunction = (previous: boolean, nextIndex: number): Q.Promise<Boolean> => {
    if (previous === exitValue) return Q(exitValue);
    if (nextIndex < conditions.length) return PromiseUtils.getPRomiseFunc(conditions[nextIndex])().then((result) => stepFunction (result, nextIndex + 1));
    return Q(exitValue);
}
return stepFunction(!exitValue, 0);
```

but I don't think it's worthwhile; the overhead in stepping through the list is most likely negligible.

src/taco-utils/promisesUtils.ts
nisheetjain: done

src/taco-utils/promisesUtils.ts
nisheetjain: fixed

PR #201, opened 2015-12-08
Adding mocha.opts for TACO tests. it defaults our timeout to 150s

nisheetjain:
This way we can quickly tweak this value to check if tests pass at least
functionally


MSLaguana:
Looks good to me

PR #202, opened 2015-12-09
Added NpmHelper to handle invoking NPM commands programmatically

joshuaskelton:
## Summary

Refactoring our interactions with npm to a programmatic one. The invocation of npm commands is handled by the NpmHelper static class. The changes are modeled after how Cordova interacts with npm.

## Changes
- Created helper class NpmHelper.
- Updated the validateCliVersion() for kits to use NpmHelper.
- Updated tacoPackageLoader to use NpmHelper.
- Added typings for npm package.


src/taco-cli/cli/kit.ts
```diff
@@ -664,27 +666,26 @@ class Kit extends commands.TacoCommandBase {
...
-        CordovaHelper.ensureCordovaVersionAcceptable(version);
 
-        var npmCommand: string = process.platform === "win32" ? "npm.cmd" : "npm";
-        var npmProcess: child_process.ChildProcess = child_process.spawn(npmCommand, ["view", "cordova", "versions"]);
-        npmProcess.on("error", function (error: Error): void {
-            throw errorHelper.get(TacoErrorCodes.ErrorReadingPackageVersions, "cordova");
-        });
+        CordovaHelper.ensureCordovaVersionAcceptable(version);
+        
+        Q({})
```
nisheetjain: we don't need Q({}) here. you can directly call NpmHelper.view

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+				if (optionalArgs !== undefined) {
+					return Q.ninvoke(npm.commands, npmCommand, args, optionalArgs);
+				}
+				else {
+					return Q.ninvoke(npm.commands, npmCommand, args);
+				}
+			});
+		}
+		
+		public static install(args: string[], path: string, logLevel?: InstallLogLevel): Q.Promise<any> {
```
nisheetjain: NpmHelper.install should really looks like install(packageId:string, workingDirectory: string, logLevel?...)

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+			return NpmHelper.runNpmCommand("install", path, args, undefined, logLevel);
+		}
+		
+		public static view(args: string[], path: string, logLevel?: InstallLogLevel): Q.Promise<any> {
```
nisheetjain: +1

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+import npm = require("npm");
+import path = require("path");
+import installLogLevel = require("./installLogLevel");
+import Q = require("q");
+
+import InstallLogLevel = installLogLevel.InstallLogLevel;
+
+module TacoUtility {
+	export class NpmHelper {
+		private static runNpmCommand(npmCommand: string, path: string, args: string[], optionalArgs: any, logLevel?: InstallLogLevel): Q.Promise<any> {
```
MSLaguana: Since this method is private and you always call it with logLevel (possibly undefined, but provided) but not always with optionalArgs, might be worth reordering the arguments.

src/taco-utils/npmHelper.ts
MSLaguana: I'd also add, what type do you expect this to return? I know that npm.d.ts just says it calls back with an any type, but it looks like the caller of this function assumes that it returns a dictionary of some form?

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+			return Q.nfcall(npm.load, {}).then(function() {
```
MSLaguana: So in the end we decided against having a single npm.load and went with a load-per-operation? Is that because load is a no-op for npm after the first time?

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+				if (npm.config.list[0]) {
+					npm.config.list[0]["logLevel"] = InstallLogLevel[logLevel];
```
MSLaguana: Should we also be checking if `logLevel` is defined here?

src/taco-utils/package.json
```diff
@@ -35,7 +35,9 @@
         "semver": "2.3.1",
         "unorm": "1.3.3",
         "windows-no-runnable": "0.0.6",
-        "wrench": "^1.5.8"
+        "wrench": "^1.5.8",
+        "lodash": "^3.10.1",
```
MSLaguana: Where is this used? I don't see it mentioned in this PR

src/taco-cli/cli/kit.ts
```diff
@@ -664,27 +666,26 @@ class Kit extends commands.TacoCommandBase {
...
+                    versions = result[cordovaVersion].versions;
+                }
 
-        npmProcess.stdout.on("data", function (data: any): void {
-            try {
-                // The versions are returned as stringified array of strings
-                var versions: string = data.toString();
-                if (versions.indexOf("'" + version + "'") !== -1) {
+                if (versions.indexOf(version) !== -1) {
                     deferred.resolve(version);
```
MSLaguana: Now that everything is promisey, we don't need to use a deferred any more, you can just `return NpmHelper.view(...).then(...).then(() => {... return version })`

src/taco-cli/cli/kit.ts
```diff
@@ -664,27 +666,26 @@ class Kit extends commands.TacoCommandBase {
         if (!semver.valid(version)) {
```
MSLaguana: Not directly related, but while looking at `npm view` I saw that we could also look at `npm view cordova dist-tags` to get the tags which are not semver valid. Might be worth in future changing the logic here to just invoke `npm view cordova dist-tags versions` and checking to see if the provided value turns up anywhere in there

src/taco-utils/package.json
joshuaskelton: This might be an incorrect merge.

joshuaskelton: It looks like lodash is used by telemetry, telemetryHelper, and promisesUtils. Wrench is used by projectHelper.

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,89 @@
...
+					// Set new value from config object
+					npm.config.list[0][propertiesToSet[i]] = conf[propertiesToSet[i]];
+				}
+
+				npm.prefix = workingDirectory || ".";
+
+				if (optionalArgs !== undefined) {
+					return Q.ninvoke(npm.commands, npmCommand, args, optionalArgs);
+				}
+				else {
```
nisheetjain: Nit: else not needed

src/taco-cli/cli/kit.ts
nisheetjain: not your code but we can resolve {} here since we don't use the resolved version anyways

src/taco-cli/cli/kit.ts
```diff
@@ -664,27 +666,23 @@ class Kit extends commands.TacoCommandBase {
...
 
-        var npmCommand: string = process.platform === "win32" ? "npm.cmd" : "npm";
-        var npmProcess: child_process.ChildProcess = child_process.spawn(npmCommand, ["view", "cordova", "versions"]);
-        npmProcess.on("error", function (error: Error): void {
-            throw errorHelper.get(TacoErrorCodes.ErrorReadingPackageVersions, "cordova");
-        });
+        CordovaHelper.ensureCordovaVersionAcceptable(version);
+        
+        NpmHelper.view("cordova", ["versions"])
+            .then(function(result: any) {
```
nisheetjain: does this have to be any?

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,89 @@
...
+var configDefs = npmconf.defs;
+var types = configDefs.types;
+var shorthands = configDefs.shorthands;
+
+import installLogLevel = require("./installLogLevel");
+import InstallLogLevel = installLogLevel.InstallLogLevel;
+
+module TacoUtility {
+	export class NpmHelper {
+		private static runNpmCommand(npmCommand: string, args: string[], workingDirectory: string, commandFlags?: string[], logLevel?: InstallLogLevel, optionalArgs?: any): Q.Promise<any> {
```
nisheetjain: are we using optionalArgs?

src/typings/npmHelper.d.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+/// <reference path="../typings/installLogLevel.d.ts" />
+/// <reference path="../typings/q.d.ts" />
+
+declare module TacoUtility {
+    class NpmHelper {
+        public static install(packageId: string, workingDirectory?: string, commandFlags?: string[], logLevel?: InstallLogLevel): Q.Promise<any>;
+
+		public static view(packageId: string, fields?: string[], workingDirectory?: string, commandFlags?: string[], logLevel?: InstallLogLevel): Q.Promise<any>;
```
nisheetjain: is the space off here for view?

src/taco-utils/npmHelper.ts
```diff
@@ -0,0 +1,89 @@
...
+			}
+			
+			var oldValues: any = {};
+			return Q.nfcall(npm.load, {}).then(function() {
+				for (var i: number = 0; i < propertiesToSet.length; i++) {
+					// Record current value
+					oldValues[propertiesToSet[i]] = Object.keys(npm.config.list[0]).indexOf(propertiesToSet[i]) === -1 ? undefined : npm.config.list[0][propertiesToSet[i]];
+					// Set new value from config object
+					npm.config.list[0][propertiesToSet[i]] = conf[propertiesToSet[i]];
+				}
```
nisheetjain: can these 2 for loops be simplified as following?

var configList = npm.config.list[0];
propertiesToSet
.foreach(p => {
   oldValues[p] = !!configList[p] ? configList[p]: undefined;
   configList[p] = conf[p];
}

propertiesToSet.foreach(p => {
   if (oldValues[p]) {
      configList[p] = oldValues[p];
   } else {
      delete configList[p];
   }
}

src/taco-utils/npmHelper.ts
joshuaskelton: Yes, view uses it to suppress console output. Some npm commands use more parameters than their signature would suggest, and this is a way to make use of that.

src/typings/npmHelper.d.ts
joshuaskelton: Yeah, looks like I never configured my new install of VSCode...

src/taco-utils/npmHelper.ts
joshuaskelton: Fixed.

src/taco-cli/cli/kit.ts
joshuaskelton: Nope.

src/taco-utils/npmHelper.ts
joshuaskelton: That is a good simplification. Code updated.

nisheetjain:
:+1:

PR #203, opened 2015-12-09
Telemetry fixes

meenakb:
Fixing three issues 
1. TACO hangs on VS dev comand prompts 
2. Response to telemetry prompt is printed twice on Mac
3. Machine ID never read from the registry on Mac

### DIFF
--- a/src/taco-utils/telemetry.ts
+++ b/src/taco-utils/telemetry.ts
@@ -26,6 +26,7 @@ import readline = require ("readline");
 import sender = require ("applicationinsights/Library/Sender");
 import telemetryLogger = require ("applicationinsights/Library/Logging");
 import winreg = require("winreg");
+
 import utilHelper = require ("./utilHelper");
 import utilResources = require ("./resources/resourceManager");
 
@@ -350,14 +351,16 @@ module TacoUtility {
             private static getRegistryValue(key: string, value: string, hive: winreg.Hive): Q.Promise<string> {
                 var deferred: Q.Deferred<string> = Q.defer<string>();
                 var regKey = new winreg({
-                                        hive: hive,                                          // HKEY_CURRENT_USER
-                                        key:  key // key containing autostart programs
+                                        hive: hive,
+                                        key:  key
                                 });
                 regKey.get(value, function (err: any, itemValue: winreg.WinregValue) {
-                    if (err)
+                    if (err) {
                         deferred.reject(err);
-                    else
+                    }
+                    else {
                         deferred.resolve(itemValue.value);
+                    }
                 });
 
                 return deferred.promise;
@@ -389,22 +392,18 @@ module TacoUtility {
                 var uniqueId: string;
                 var deferred: Q.Deferred<string> = Q.defer<string>();
                 if (os.platform() === "win32") {
-                    TelemetryUtils.getRegistryValue(TelemetryUtils.REGISTRY_SQMCLIENT_NODE, regValue, regHive)
-                    .then(function(id: string): void {
+                    return TelemetryUtils.getRegistryValue(TelemetryUtils.REGISTRY_SQMCLIENT_NODE, regValue, regHive)
+                    .then(function(id: string): Q.Promise<string> {
                         if (id) {
                             uniqueId = id.replace(/[{}]/g, "");
+                            return Q.resolve(uniqueId);
                         } else {
-                            uniqueId = fallback();
+                            return Q.resolve(fallback());
                         }
-
-                        deferred.resolve(id);
                     });
                 } else {
-                    uniqueId = fallback();
-                    deferred.resolve(uniqueId);
+                    return Q.resolve(fallback());
                 }
-                
-                return deferred.promise;
             }
 
             private static generateMachineId(): string {
--- a/src/taco-utils/telemetry.ts
+++ b/src/taco-utils/telemetry.ts
@@ -10,6 +10,7 @@
 
 /// <reference path="../typings/node.d.ts" />
 /// <reference path="../typings/applicationinsights.d.ts" />
+/// <reference path="../typings/winreg.d.ts" />
 
 import appInsights = require ("applicationinsights");
 import crypto = require ("crypto");
@@ -24,6 +25,7 @@ import Q = require ("q");
 import readline = require ("readline");
 import sender = require ("applicationinsights/Library/Sender");
 import telemetryLogger = require ("applicationinsights/Library/Logging");
+import winreg = require("winreg");
 import utilHelper = require ("./utilHelper");
 import utilResources = require ("./resources/resourceManager");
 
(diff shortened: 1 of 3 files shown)


src/taco-utils/telemetry.ts
```diff
@@ -293,7 +296,7 @@ module TacoUtility {
                 logger.logLine();
 
                 var deferred: Q.Deferred<boolean> = Q.defer<boolean>();
-                var yesOrNoHandler = readline.createInterface({ input: process.stdin, output: process.stdout });
+                var yesOrNoHandler = readline.createInterface({ input: process.stdin, output: process.stdout, terminal: false });
```
MSLaguana: Just to be sure: We deliberately don't want to print out the values used when doing something like `taco feedback < answersFile`?

src/taco-utils/telemetry.ts
```diff
@@ -344,17 +347,20 @@ module TacoUtility {
...
-                    if (regKey && regKey[value] && regKey[value].value) {
-                        return regKey[value].value;
-                    }
-                } catch (e) {
-                    return null;
-                }
+            private static getRegistryValue(key: string, value: string, hive: winreg.Hive): Q.Promise<string> {
+                var deferred: Q.Deferred<string> = Q.defer<string>();
+                var regKey = new winreg({
+                                        hive: hive,                                          // HKEY_CURRENT_USER
```
MSLaguana: Are these comments accurate?

src/taco-utils/telemetry.ts
```diff
@@ -344,17 +347,20 @@ module TacoUtility {
...
+                                        key:  key // key containing autostart programs
+                                });
+                regKey.get(value, function (err: any, itemValue: winreg.WinregValue) {
+                    if (err)
```
MSLaguana: Braces for if statements, or it's begging for future mistakes :)

src/taco-utils/telemetry.ts
```diff
@@ -379,15 +385,45 @@ module TacoUtility {
                 fs.writeFileSync(TelemetryUtils.telemetrySettingsFile, JSON.stringify(TelemetryUtils.telemetrySettings));
             }
 
-            private static getMachineId(): string {
+            private static getUniqueId(regValue: string, regHive: winreg.Hive, fallback: () => string): Q.Promise<any> {
+                var uniqueId: string;
+                var deferred: Q.Deferred<string> = Q.defer<string>();
+                if (os.platform() === "win32") {
+                    TelemetryUtils.getRegistryValue(TelemetryUtils.REGISTRY_SQMCLIENT_NODE, regValue, regHive)
```
MSLaguana: Rather than using deferreds, you can just return the promise chain here, and in the else branch you can return `Q.resolve(fallback())`

src/taco-utils/telemetry.ts
meenakb: Correct, confirmed this with Nisheet as well.

meenakb:
All comments have been addressed.

MSLaguana:
looks good to me

src/taco-utils/telemetry.ts
```diff
@@ -225,23 +227,25 @@ module TacoUtility {
...
-                TelemetryUtils.machineId = TelemetryUtils.getMachineId();
-                TelemetryUtils.sessionId = TelemetryUtils.generateGuid();
-                TelemetryUtils.userType = TelemetryUtils.getUserType();
-                
-
-                if (_.isUndefined(isOptedInValue)) {
-                    return TelemetryUtils.getOptIn()
-                    .then(function (optIn: boolean): void {
-                        Telemetry.isOptedIn = optIn;
+                return Q.all([TelemetryUtils.getUserId(), TelemetryUtils.getMachineId()])
```
nisheetjain: shouldn't we get UserId/machineId etc only if user has opted in?

src/taco-utils/telemetry.ts
```diff
@@ -344,17 +348,22 @@ module TacoUtility {
...
-                        return regKey[value].value;
+            private static getRegistryValue(key: string, value: string, hive: winreg.Hive): Q.Promise<string> {
+                var deferred: Q.Deferred<string> = Q.defer<string>();
+                var regKey = new winreg({
+                                        hive: hive,
+                                        key:  key
+                                });
+                regKey.get(value, function (err: any, itemValue: winreg.WinregValue) {
+                    if (err) {
+                        deferred.reject(err);
```
nisheetjain: what is the end user behavior if regKey.get fails?

src/taco-utils/telemetry.ts
```diff
@@ -379,15 +388,41 @@ module TacoUtility {
...
-            private static getMachineId(): string {
+            private static getUniqueId(regValue: string, regHive: winreg.Hive, fallback: () => string): Q.Promise<any> {
+                var uniqueId: string;
+                var deferred: Q.Deferred<string> = Q.defer<string>();
+                if (os.platform() === "win32") {
+                    return TelemetryUtils.getRegistryValue(TelemetryUtils.REGISTRY_SQMCLIENT_NODE, regValue, regHive)
+                    .then(function(id: string): Q.Promise<string> {
+                        if (id) {
+                            uniqueId = id.replace(/[{}]/g, "");
+                            return Q.resolve(uniqueId);
```
nisheetjain: Nit: you can use Q(uniqueId) as well also else is unnecessary

src/taco-utils/telemetry.ts
```diff
@@ -379,15 +388,41 @@ module TacoUtility {
...
+            }
+
+            private static getMachineId(): Q.Promise<string> {
                 var machineId: string = TelemetryUtils.telemetrySettings.machineId;
                 if (!machineId) {
-                    var macAddress: string = TelemetryUtils.getMacAddress();
-                    machineId = crypto.createHash("sha256").update(macAddress, "utf8").digest("hex");
+                    return TelemetryUtils.getUniqueId(TelemetryUtils.REGISTRY_MACHINEID_VALUE, winreg.HKLM, TelemetryUtils.generateMachineId)
+                    .then(function(id: string): Q.Promise<string> {      
+                        TelemetryUtils.telemetrySettings.machineId = id;
```
nisheetjain: why are we setting all properties twice? once TelemetryUtils.machineId = id and also TelemetryUtils.telemetrySettings.machineId

src/taco-utils/telemetry.ts
meenakb: We have been writing {"userId":"X-X","machineId":"Y-Y","userType":"Z","optIn":false} even if the user opts out. I do see that it seems a bit unnecessary. What do you suggest? We track the current behavior as an issue if we decide to.

src/taco-utils/telemetry.ts
meenakb: Good catch, that has been fixed.

src/taco-utils/telemetry.ts
meenakb: ah, correct - fixing.

PR #205, opened 2015-12-10
Fail gracefully by returning null if there was an error in attempting to read from registry

meenakb:


MSLaguana:
I suggest that we also improve taco.ts to check for errors as well; have the promise chain in the main `run()` function end with a done() so we don't silently swallow errors

src/taco-cli/cli/taco.ts
```diff
@@ -107,12 +107,12 @@ class Taco {
                             });
```
MSLaguana: We'll also have to make sure to return the Taco.runWithParsedArgs promise above now

meenakb:
I agree, improving taco.ts to error out in case of any error in the entire promise chain.

meenakb:
Comments addressed.

MSLaguana:
Looks good to me now

PR #206, opened 2015-12-10
Fixing postinstall script to use exit codes instead of callbacks.

nisheetjain:
Also modifying the success/failure messages


MSLaguana:
Looks good to me

PR #207, opened 2015-12-10
Modifying platform/plugin command to support subCommands.

nisheetjain:
As part of the change also moving some common code into the base class (renamed to kitComponentCommand)
Removing all non-cordova inteface from cordovaExtensions.d.ts


src/taco-utils/commands.ts
```diff
@@ -99,11 +99,7 @@ module TacoUtility {
                 // Determine which subcommand we are executing
                 var subCommand = this.getSubCommand(this.data);
                 if (subCommand) {
-                    return subCommand.run(this.data)
-                        .then((telemetryProperties: telemetryHelper.ICommandTelemetryProperties) => {
-                            telemetryProperties["subCommand"] = telemetryHelper.TelemetryHelper.telemetryProperty(subCommand.name, /*isPii*/ false);
```
nisheetjain: this is removed to handle fallback case. we want to report the actual subcommand instead of "fallback"

src/taco-cli/cli/kit.ts
```diff
@@ -36,6 +36,8 @@ import CordovaHelper = tacoUtility.CordovaHelper;
 import logger = tacoUtility.Logger;
 import LoggerHelper = tacoUtility.LoggerHelper;
 import ProjectHelper = tacoUtility.ProjectHelper;
+import PromisesUtils = tacoUtility.PromisesUtils;
```
meenakb: Not a part of this PR, but how about calling this just "promiseUtils"?

src/taco-cli/cli/plugin.ts
```diff
@@ -9,154 +9,157 @@
...
-            return Q({});
-        }
+class Plugin extends kitComponentCommand.KitComponentCommand {
+    public name: string = "platform";
+
+    private static KNOWN_OPTIONS: Nopt.CommandData = {
+        searchpath: String,
+        noregistry: String,
+        variable: Array,
+        browserify: Boolean,
```
meenakb: Add this experimental fla "browserify" to commands.json as well.

src/taco-cli/cli/utils/kitComponentCommand.ts
```diff
@@ -0,0 +1,303 @@
...
+    spec?: string;
+    targetId: string;
+}
+
+/**
+ * KitComponentCommand
+ *
+ * Base class to handle add/remove operations for cordova components (platforms, plugins)
+ */
+export class KitComponentCommand extends commands.TacoCommandBase {
```
meenakb: Maybe CordovaComponentCommand? KitComponentCommand is a bit misleading given that this has code path for CLI projects as well.

meenakb:
:+1:

src/taco-cli/cli/plugin.ts
nisheetjain: done

src/taco-cli/cli/utils/kitComponentCommand.ts
nisheetjain: done

PR #210, opened 2015-12-10
Cordova gate

leotlee:
Enable ability to gate check new versions of Cordova by overriding the version of Cordova used in LazyRequire 

"gulp run-tests --testCordova 5.4.1"

Details:
1. put "TestCordovaExempt" in scenario name to exempt from gate check
2. exempt text is inserted into args to mocha.opts
i.e. --invert --grep TestCordovaExempt
3. version of Cordova used is stored in process.env.testCordovaVersion


src/mocha.opts
```diff
@@ -1 +1 @@
---timeout 150000
+--timeout 150000
```
MSLaguana: Is this intentional?

leotlee: Yes, there was a newline at end of file, I needed to append to it in same line.

tools/GulpUtils.ts
```diff
@@ -15,13 +15,29 @@ import Q = require ("q");
...
             args = ["--reporter", reporter];
         }
-      return GulpUtils.runNpmScript(modulesToTest, modulesRoot, GulpUtils.testCommand, failTestsAtEnd, args);
+
+        return GulpUtils.chainAsync<string>(modulesToTest, (moduleName: string) => {
+
+            if (testCordova) {
+                var mochaOptFile: string = path.resolve(modulesRoot, moduleName, GulpUtils.testDir, GulpUtils.mochaOptFile);
+                fs.appendFileSync(mochaOptFile, " " + GulpUtils.mochaTestInvert + " " + GulpUtils.testCordovaExemptText); // i.e. run all tests except for TestCordovaExempt
+                process.env.testCordovaVersion = testCordova;
```
MSLaguana: Should this assignment to process.env.testCordovaVersion just happen once? And then we could `return Q.denodeify(fs.appendFile)(mochaOptFile, ...);` in the case that we modify things rather than always returning dummy promises.

tools/GulpUtils.ts
```diff
@@ -15,13 +15,29 @@ import Q = require ("q");
 
 class GulpUtils {
     private static testCommand: string = "test";
+    private static testDir: string = "test"; // test directory
+    private static mochaOptFile: string = "mocha.opts"; // mocha config file
+    private static testCordovaExemptText: string = "TestCordovaExempt"; // test directory
```
MSLaguana: Incorrect comment

src/taco-utils/tacoPackageLoader.ts
```diff
@@ -266,6 +266,10 @@ module TacoUtility {
         private static lazyRequireInternal<T>(request: IPackageInstallRequest): Q.Promise<T> {
             assert.notEqual(request, null);
 
+            if (process.env.testCordovaVersion && request.packageName.toLocaleLowerCase() === "cordova") {
```
nisheetjain: I would really prefer if we do this in CordovaWrapper instead

tools/GulpUtils.ts
leotlee: Fixed

tools/GulpUtils.ts
leotlee: Done.

src/taco-utils/tacoPackageLoader.ts
leotlee: Addeded LazyCordovaRequire and LzyCordovaRun as discussed.

MSLaguana:
Looks good to me

PR #211, opened 2015-12-11
TACO platform list should include remote platforms as well

nisheetjain:


src/taco-cli/cli/platform.ts
```diff
@@ -170,6 +176,22 @@ class Platform extends KitComponentCommand {
...
+                // The user has invoked 'taco platform list'
+                // We want to add an additional line including any remote platforms
+                return Settings.loadSettingsOrReturnEmpty()
+                    .then((settings: Settings.ISettings): void => {
+                        var remotePlatforms: string[] = Object.keys(settings.remotePlatforms || {});
+                        if (remotePlatforms.length > 0) {
+                            Logger.log(resources.getString("PlatformListRemotePlatforms", remotePlatforms.join(",")));
+                        }
+                    })
+                    .then(() => telemetryProperties);
```
MSLaguana: Should we update the telemetry properties with anything from here? You can also just return telemetryProperties in the previous then rather than chaining a new one.

src/taco-cli/cli/platform.ts
```diff
@@ -170,6 +176,22 @@ class Platform extends KitComponentCommand {
                 break;
 
         }
+
+    }
+    private list(): Q.Promise<ICommandTelemetryProperties> {
+        return this.passthrough()
```
MSLaguana: Potentially we could kick off the settings loading as well to reduce load times?
`return Q.all([this.passthrough(), Settings.loadSettingsOrReturnEmpty()]).spread(function (telemetryProperties, settings) { ...`

src/taco-cli/cli/platform.ts
```diff
@@ -170,6 +176,22 @@ class Platform extends KitComponentCommand {
...
+            .then(function(telemetryProperties: ICommandTelemetryProperties): Q.Promise<ICommandTelemetryProperties> {
+                // The user has invoked 'taco platform list'
+                // We want to add an additional line including any remote platforms
+                return Settings.loadSettingsOrReturnEmpty()
+                    .then((settings: Settings.ISettings): void => {
+                        var remotePlatforms: string[] = Object.keys(settings.remotePlatforms || {});
+                        if (remotePlatforms.length > 0) {
+                            Logger.log(resources.getString("PlatformListRemotePlatforms", remotePlatforms.join(",")));
```
MSLaguana: Cordova will print out `Installed platforms: ` even if there aren't any. I think it might be a good idea to follow along, so the output is predictable.

nisheetjain:
Fix manually merged into master

PR #212, opened 2015-12-12
Update templates

guillaumejenkins:
Integrating asset changes from the latest VS templates


MSLaguana:
Looks good to me

MSLaguana:
Do we want to be proactive about upcoming cordova-ios platform changes? They will be adding in a new icon size for ipad pro: https://issues.apache.org/jira/browse/CB-10184

templates/default/blank/www/images/.taco-ignore
```diff
@@ -1,3 +0,0 @@
-This is a dummy file with the sole purpose of allowing to commit its parent directory (which would otherwise be empty) to git source control.
-
```
meenakb: Is this file no longer needed?

templates/default/typescript/www/images/.taco-ignore
```diff
@@ -1,3 +0,0 @@
-This is a dummy file with the sole purpose of allowing to commit its parent directory (which would otherwise be empty) to git source control.
-
-It is not intended to be part of the template and will be excluded from the archive process (when doing "gulp prepare-templates").
```
meenakb: Looks like images folder is no longer (otherwise) empty and will contain cordova.png. In that case, do you want to remove the ".taco-ignore" file type from the ignored file lists in utils\templateManager.ts?

guillaumejenkins: We discussed this offline, but putting answer here for others: we still need need the taco-ignore file for the www/scripts/ folder of the TypeScript template

meenakb:
:+1:

PR #213, opened 2015-12-14
Fix "remotebuild saveconfig" crash if taco_home doesn't exist

guillaumejenkins:


nisheetjain:
:+1:

src/remotebuild/lib/cli.ts
```diff
@@ -69,7 +69,13 @@ class CliHelper {
         nconf.argv();
         // Default to using TACO_HOME/RemoteBuild.config
         // If that file doesn't exist, then this will be equivalent to nconf.use("memory") as long as we don't try to save it out
-        var configFile: string = nconf.get("config") || path.join(UtilHelper.tacoHome, "RemoteBuild.config");
+        var configFile: string = nconf.get("config");
+
+        if (!configFile) {
+            configFile = path.join(UtilHelper.tacoHome, "RemoteBuild.config");
+            UtilHelper.createDirectoryIfNecessary(UtilHelper.tacoHome);
```
MSLaguana: If the user specifies `--config=~/taco/myconfig.config` and ~/taco doesn't yet exist, do we want to crash in that case still?

guillaumejenkins: Made changes to deal with such cases

MSLaguana:
Looks good to me

joshuaskelton:
:+1:

PR #215, opened 2015-12-15
'taco build ios --clean' no longer sets the --release flag by mistake

guillaumejenkins:


MSLaguana:
This looks correct to me now, but I think that my original choice to modify the commandData flags is a poor one. Considering we don't actually use them apart from deciding which folders to clean out, might be worth constructing that list separately, maybe even declaring commandData as const (if you can even do that with function arguments)

MSLaguana:
Looks good to me

joshuaskelton:
:+1:

PR #216, opened 2015-12-15
Fixing an issue with a kit update test. As part of this, initializing the download options to defaults expected by Cordova 

meenakb:


src/taco-cli/cli/kit.ts
```diff
@@ -427,7 +427,7 @@ class Kit extends commands.TacoCommandBase {
             logger.log(resources.getString("CommandKitSelectStatusUpdatingPlugins"));
         }
 
-        var downloadOptions: Cordova.ICordovaDownloadOptions = { searchpath: "", noregistry: false, usegit: false, cli_variables: {}, browserify: "", link: "", save: true, shrinkwrap: false };
+        var downloadOptions: Cordova.ICordovaDownloadOptions = { searchpath: undefined, noregistry: undefined, usegit: undefined, cli_variables: {}, browserify: undefined, link: undefined, save: true, shrinkwrap: undefined };
```
MSLaguana: Perhaps we should modify our typings here and make all these undefined entries optional, and simply not specify them here? An object that sets `foo: undefined` is subtly different to an object that does not define `foo`.

meenakb: Good point, I see some changes in the interface as part of Nisheet's PR. I will wait until he merges.

src/taco-cli/test/kit.ts
```diff
@@ -243,23 +243,23 @@ describe("Kit command: ", function (): void {
...
-                // change the encoding to UTF 16
-                var tacoJsonContents: string = fs.readFileSync(tacoJsonPath).toString();
-                fs.writeFileSync(tacoJsonPath, new Buffer(tacoJsonContents, "utf16le"));
-                return kitRun(["list"]);
-            })
-            .then((telemetryParameters: TacoUtility.ICommandTelemetryProperties) => {
-                var expected: TacoUtility.ICommandTelemetryProperties = { subCommand: { isPii: false, value: "list" } };
-                telemetryParameters.should.be.eql(expected);
-            });
+        it("'taco kit list should work for a project UTF16 taco.json", function (): Q.Promise<any> {
```
nisheetjain: what has changed here?

nisheetjain:
rest all LG

src/taco-cli/test/kit.ts
meenakb: The test "taco kit list should work for a project UTF16 taco.json" is now part of the parent suite. This improves the readability of test results.

PR #221, opened 2015-12-16
Refactoring live reload into taco-cli

nisheetjain:
This would reduce test run time and overall performance


nisheetjain:
@omefire can you please take a look

src/taco-cli/cli/run.ts
```diff
@@ -167,21 +161,31 @@ class Run extends commands.TacoCommandBase {
         });
     }
 
+    private runLocal(localPlatforms?: string[]): Q.Promise<tacoUtility.ICommandTelemetryProperties> {
+        if (this.data.options["livereload"] || this.data.options["devicesync"]) {
+            // intentionally delay-requiring it since liveReload fetches whole bunch of stuff
+            return require("./liveReload").startLiveReload(!!this.data.options["livereload"], !!this.data.options["devicesync"], localPlatforms);
```
omefire: It seems to me like we need NOT return here, since we have just set event handlers for livereload, we haven't really "ran" anything yet.

omefire: Ooh wait, scratch that.
Just looking at this line: https://github.com/Microsoft/TACO/blob/b1de83577c878bf81774ba34389514f8efc656a1/src/taco-cli/cli/liveReload.ts#L33,

I realize that we use Cordova.tryInvoke(...)

src/taco-cli/cli/liveReload.ts
```diff
@@ -0,0 +1,97 @@
...
+                            Logger.log("Press CTRL+C to exit");
+                        }
+                    });
+
+                    cordova.on("after_prepare", function() {
+                        if (!liveReload.isLiveReloadActive()) {
+                            return liveReload.start(projectRoot, targetPlatforms, options);
+                        }
+                    });
+                }, () => Logger.log("Unsupported for non-taco projects "), {});
```
omefire: :nitpicking here but maybe, let's change this message to "Livereload is not supported for projects that were not created with TACO" ?
Just to clarify things...

omefire:
LGTM! ... except for minor details.

omefire:
Hmm, there is a case where the user doesn't have the '--remote' flag issued at the command line, but has a remote platform configured. it is my understanding that in that case, we run the platform as remote.
If my understanding is correct, the parseArgs wouldn't catch this error case, and we'll need an additional check.

Could you please double-check that ?

nisheetjain:
@omefire yes you are right. parseArg wouldn't handle that case. I will add the check back.

omefire:
Also, let's add a comment about ```parseArgs``` not being enough to handle that case so that it doesn't mistakenly get refactored away in the future.

src/taco-cli/cli/run.ts
omefire: Hmm, looking at this closely, we'll still need to issue a ```cordova.raw.run(...)``` command after ```startLiveReload```.

src/taco-cli/cli/liveReload.ts
```diff
@@ -0,0 +1,97 @@
...
+                    };
+
+                    liveReload.on("livereload:error", function(err) {
+                        Logger.log("TACO Livereload Error occured: - " + err);
+                    });
+            
+                    // Display prompt to the user letting them know that livereload has started
+                    // Note: we do it this way because doing it after `lr.start()`, it would get lost
+                    //          among all the run command details output.
+                    cordova.on("after_run", function() {
```
omefire: We'll also need an event handler for the "emulate" case. I missed this the first time I was working on the feature.

nisheetjain: can you fix this as a bugfix? I have committed this PR into master

PR #222, opened 2015-12-16
Remotebuild: Handle client Cordova version changes for incremental builds

guillaumejenkins:
~~This is an early pull request for early feedback, I haven't tested thoroughly enough~~

Validation done, this is working well


src/taco-remote-lib/common/builder.ts
```diff
@@ -128,10 +129,89 @@ class Builder {
...
+                return this.update_platform();
+            });
         }
     }
 
+    /**
+     * Removes and add the platform if necessary (Cordova version changed, or config.xml has a saved platform version that differs from the installed one). Returns true if the platform was added,
+     * false if not.
+     */
+    private reAddPlatformIfNecessary(): Q.Promise<boolean> {
```
nisheetjain: how about we rename reAddPlatformIfNecessary to addPlatform or ensurePlatformAdded
addPlatform can also have the !fs.existsSync case inside it

AddPlatform logic could be
if (!exist) {add} else if (versionmismatch) {remove; add}

src/taco-remote-lib/common/builder.ts
```diff
@@ -123,13 +124,98 @@ class Builder {
...
+            return this.cordova.raw.platform("add", this.currentBuild.buildPlatform).then(() => {
+                return Q(true);
+            });
+        }
+
+        var xmlVersion: string = this.getConfigXmlPlatformVersion();
+
+        return this.getCurrentPlatformVersion().then((installedVersion: string) => {
+            if (xmlVersion) {
+                if (semver.neq(xmlVersion, installedVersion)) {
```
MSLaguana: I think this is wrong. I think you want to use `semver.satisfies` to let it match ranges, and it looks like it expects `semver.satisfies(singleVersion, versionSpecifier)`. As it stands, if `xmlVersion` is something like `~1.1.0` it'll complain.

src/taco-remote-lib/common/builder.ts
```diff
@@ -123,13 +124,98 @@ class Builder {
...
-            return this.update_platform();
+            return this.cordova.raw.platform("add", this.currentBuild.buildPlatform).then(() => {
+                return Q(true);
+            });
+        }
+
+        var xmlVersion: string = this.getConfigXmlPlatformVersion();
+
+        return this.getCurrentPlatformVersion().then((installedVersion: string) => {
+            if (xmlVersion) {
```
MSLaguana: If we only consider `installedVersion` if `xmlVersion` exists, should we only invoke `this.getCurrentPlatformVersion()` if `xmlVersion` is set?

src/taco-remote-lib/common/builder.ts
```diff
@@ -123,13 +124,98 @@ class Builder {
...
+
+        return "";
+    }
+
+    private getCurrentPlatformVersion(): Q.Promise<string> {
+        var deferred: Q.Deferred<string> = Q.defer<string>();
+
+        var resolvePlatformOutput: Function = (output: string) => {
+            var versionRegexp: RegExp = new RegExp(util.format("%s (.*?)(?:,|$)", this.currentBuild.buildPlatform), "m");
+            var installedPlatformVersion: string = versionRegexp.exec(output)[1];
```
MSLaguana: Is it possible that the output won't match this?
Also is it necessary to use multiline matching? It should all be on the one line like `Installed Platforms: ios 3.9.2, android 5.0.1` right?

src/taco-remote-lib/common/builder.ts
```diff
@@ -123,13 +124,98 @@ class Builder {
...
+                    return this.cordova.raw.platform("add", this.currentBuild.buildPlatform);
+                }).then(() => {
+                    return Q(true);
+                });
+            }
+        });
+    }
+
+    private getConfigXmlPlatformVersion(): string {
+        var engines: {[key: string]: string} = utils.CordovaConfig.getCordovaConfig(this.currentBuild.appDir).engines();
```
MSLaguana: Do you need to specify the type here rather than relying on the .d.ts?

src/taco-remote-lib/common/builder.ts
guillaumejenkins: Will do

src/taco-remote-lib/common/builder.ts
guillaumejenkins: Will do

src/taco-remote-lib/common/builder.ts
guillaumejenkins: I'll add guards and return an empty string if there is no match.

Multiline option is needed because there is a second line indicating available paltforms

src/taco-remote-lib/common/builder.ts
guillaumejenkins: I just went with our convention of always specifying the types, but I can remove it

src/taco-remote-lib/common/builder.ts
guillaumejenkins: Actually I'll change this to execute the version script in platformRoot/cordova/version, so none of this will be needed

MSLaguana: Multiline mainly means that `.` also matches newlines (and `$` matches end of line) so that you can have a match across multiple lines; from what I understand of this regex, there's no need for it to span across lines

src/taco-remote-lib/common/builder.ts
```diff
@@ -15,11 +15,13 @@
...
 "use strict";
 
+import child_process = require("child_process");
 import fs = require ("fs");
 import path = require ("path");
 import Q = require ("q");
 import rimraf = require ("rimraf");
-import semver = require ("semver");
+import semver = require("semver");
+import util = require("util");
```
guillaumejenkins: No need for util anymore, I'll remove it

PR #224, opened 2015-12-18
Add Telemetry to Remotebuild

joshuaskelton:
## Summary

Adding basic telemetry points to remotebuild. Telemetry is captured on server start and per build.

## Changes

* Added a new telemetry settings file for remotebuild.
* Added telemetry init/prompt to server start.
* Added telemetry after build complete.

### DIFF
--- a/src/remotebuild/lib/cli.ts
+++ b/src/remotebuild/lib/cli.ts
@@ -23,6 +23,7 @@ import hostSpecifics = require ("./hostSpecifics");
 import RemoteBuildConf = require ("./remoteBuildConf");
 import resources = require ("../resources/resourceManager");
 import utils = require ("taco-utils");
+import telemetry = utils.Telemetry;
 
 import HostSpecifics = hostSpecifics.hostSpecifics;
 import Logger = utils.Logger;
@@ -38,6 +39,9 @@ class CliHelper {
             .then(function (): void {
                 // if version flag found, show version and exit
                 CliHelper.handleVersionFlag(args);
+            })
+            .then(function(): Q.Promise<any> {
+                return telemetry.init("REMOTE_BUILD", require("../package.json").version);
             })
             .then(function (): Q.Promise<void> {
                 remotebuildConf = CliHelper.parseRemoteBuildConf();
--- a/src/remotebuild/lib/cli.ts
+++ b/src/remotebuild/lib/cli.ts
@@ -41,7 +41,7 @@ class CliHelper {
                 CliHelper.handleVersionFlag(args);
             })
             .then(function(): Q.Promise<any> {
-                return telemetry.init("REMOTE_BUILD", require("../package.json").version);
+                return telemetry.init("REMOTEBUILD", require("../package.json").version, undefined ,"RemotebuildTelemetrySettings.json");
             })
             .then(function (): Q.Promise<void> {
                 remotebuildConf = CliHelper.parseRemoteBuildConf();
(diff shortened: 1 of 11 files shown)


src/remotebuild/lib/server.ts
```diff
@@ -72,21 +73,29 @@ class Server {
...
-            if (err.stack) {
-                Logger.logError(err.stack);
-            }
-
-            throw err;
-        });
+        return utils.TelemetryHelper.generate("start",
+                    (telemetry: utils.TelemetryGenerator) => {
+                        telemetry
+                            .add("remotebuildVersion", require("../package.json").version, false)
```
MSLaguana: Is this needed? Or is version done as part of telemetry.init?

src/taco-remote/lib/buildManager.ts
```diff
@@ -431,6 +432,17 @@ class BuildManager {
                 } else {
                     self.buildMetrics.failed++;
                 }
+                
+                utils.TelemetryHelper.generate("build",
```
MSLaguana: My understanding is that `telemetry.generate` is intended for timing things. Should we start the telemetry.generate outside the `buildInfo["pkg"].build` and have it time the build?

src/taco-utils/telemetry.ts
```diff
@@ -205,6 +206,10 @@ module TacoUtility {
             private static INTERNAL_USER_ENV_VAR: string = "TACOINTERNAL";
 
             private static get telemetrySettingsFile(): string {
+                if (Telemetry.appName === "REMOTE_BUILD") {
```
MSLaguana: This feels a little too tightly coupled to me. Perhaps we should have telemetry.init accept an optional settings filename, and leave TelemetrySettings.json as a default?

src/remotebuild/lib/cli.ts
```diff
@@ -39,6 +40,9 @@ class CliHelper {
                 // if version flag found, show version and exit
                 CliHelper.handleVersionFlag(args);
             })
+            .then(function(): Q.Promise<any> {
+                return telemetry.init("REMOTE_BUILD", require("../package.json").version);
```
joshuaskelton: Just a thought: Should the ID be "TACO/remotebuild"? Or should this be treated as a separate app?
@digeff

src/taco-remote/lib/buildManager.ts
joshuaskelton: The team decided that timing the build isn't a high priority, so I'm not worrying about it.

src/remotebuild/lib/server.ts
joshuaskelton: Good catch. That is captured as part of init.

MSLaguana:
Looks good to me

src/remotebuild/lib/cli.ts
```diff
@@ -39,6 +40,9 @@ class CliHelper {
                 // if version flag found, show version and exit
                 CliHelper.handleVersionFlag(args);
             })
+            .then(function(): Q.Promise<any> {
+                return telemetry.init("REMOTEBUILD", require("../package.json").version, undefined ,"RemotebuildTelemetrySettings.json");
```
digeff: I'd personally use null here instead of undefined, and change the code to make null have the proper semantics.

src/taco-remote/lib/buildManager.ts
```diff
@@ -431,6 +432,17 @@ class BuildManager {
...
+                    (telemetry: utils.TelemetryGenerator) => {
```
digeff: Do we care about getting telemetry for the cases where this method returns early? or throws an exception, or crashes?
A best practice is to generate telemetry as soon as you get the data, so in case that the code follows a different path, or has an issue, you'll still get it.
If we aren't interested in those scenarios, then we can leave it as is. If we want to capture them, we should refactor the telemetry so it's generated as soon as possible.

PR #225, opened 2015-12-18
Remote build log is now printed on the local console, and the build l…

guillaumejenkins:
…og for incremental builds is now properly updated with the output from the new build


src/taco-cli/cli/remoteBuild/remoteBuildClientHelper.ts
```diff
@@ -81,6 +82,15 @@ class RemoteBuildClientHelper {
                     }
                 }
 
+                // Clear build log if it exists
+                if (fs.existsSync(buildLog)) {
```
MSLaguana: Unrelated, but now that we have a few filesystem operations going on here, maybe we should stop them being synchronous.

MSLaguana:
Looks good to me

src/taco-cli/cli/remoteBuild/remoteBuildClientHelper.ts
```diff
@@ -67,6 +67,7 @@ class RemoteBuildClientHelper {
 
         var changeTimeFile: string = path.join(settings.platformConfigurationBldDir, "lastChangeTime.json");
         var lastChangeTime: { [file: string]: number } = {};
+        var buildLog: string = path.join(settings.platformConfigurationBldDir, "build.log");
```
MSLaguana: Perhaps extract this further out and use the same variable in logBuildOutput rather than repeating the logic for constructing it?

PR #226, opened 2015-12-22
Now we remove the plugins saved in the config.xml file if we change to a kit that overrides them

digeff:
First we show a new message to the user, saying that we are going to remove that plugin from the config.xml.
Then we remove the plugins.


src/taco-cli/cli/kit.ts
```diff
@@ -436,13 +438,15 @@ class Kit extends commands.TacoCommandBase {
...
             var pluginOpts: Cordova.ICordovaPluginOptions = { save: true };
             return CordovaWrapper.plugin("remove", null, Object.keys(components))
-                .then(function(): Q.Promise<any> {
+                .then(function (): Q.Promise<any> {
                     // Do not save in the case of updating to CLI project
                     pluginOpts.save = editParams.isKitProject;
                     return tacoUtility.PromisesUtils.chain(Object.keys(components), (componentName: string) => {
                         // No override on the case of CLI project update - Cordova CLI gets its pinned version
                         var componentOverride: string = editParams.isKitProject ? componentName + "@" + components[componentName] : componentName;
+                        // call cordova helper to read config.xml
```
MSLaguana: Is this comment correct? It looks like you're just adding the plugins, not reading config.xml?

digeff: You are right. That comment is in the wrong place. Removing it!

src/taco-cli/cli/kit.ts
```diff
@@ -510,13 +520,15 @@ class Kit extends commands.TacoCommandBase {
...
      */
     private static printProjectUpdateInfo(id: string, installedPlatformVersions: IDictionary<string>, installedPluginVersions: IDictionary<string>,
-        platformVersionUpdates: IDictionary<string> = null, pluginVersionUpdates: IDictionary<string> = null): void {
+        platformVersionUpdates: IDictionary<string> = null, pluginVersionUpdates: IDictionary<string> = null,
+        pluginsInConfigVersions: IDictionary<string> = null, pluginsInConfigUpdates: IDictionary<string> = null): void {
         var indent: number = LoggerHelper.getDescriptionColumnIndent(Kit.getLongestPlatformPluginLength(installedPlatformVersions ? Object.keys(installedPlatformVersions) : null, installedPluginVersions ? Object.keys(installedPluginVersions) : null));
 
         var platformsRequireUpdate: boolean = Kit.projectComponentNeedsUpdate(installedPlatformVersions, platformVersionUpdates);
         var pluginsRequireUpdate: boolean = Kit.projectComponentNeedsUpdate(installedPluginVersions, pluginVersionUpdates);
+        var removesPluginsFromConfig: boolean = Kit.projectComponentNeedsUpdate(pluginsInConfigVersions, pluginsInConfigUpdates);
```
MSLaguana: should this variable be `removedPluginsFromConfig`?

src/taco-cli/cli/kit.ts
```diff
@@ -452,12 +456,18 @@ class Kit extends commands.TacoCommandBase {
...
+        installedPluginVersions: IDictionary<string>, platformVersionUpdates: IDictionary<string> = null,
+        pluginVersionUpdates: IDictionary<string> = null, savedPluginVersionUpdates: IDictionary<string> = null): Q.Promise<any> {
         logger.log(resources.getString("CommandKitSelectStatusUpdatingTacoJson"));
         // First, update the taco.json to reflect the new CLI/kit version
         // Then go ahead and update the platforms and plugins
         return kitHelper.editTacoJsonFile(editParams)
+        .then(() => ProjectHelper.getProjectInfo())
+        .then(projectInfo => {
+            var plugins: Cordova.ICordovaPluginInfo[] = Object.keys(savedPluginVersionUpdates).map(pluginName => { return { name: pluginName, spec: null }; });
+            return CordovaHelper.editPluginVersionSpecs(plugins, projectInfo);
```
meenakb: Add some comments around this.

src/taco-utils/cordovaHelper.ts
```diff
@@ -341,6 +341,14 @@ module TacoUtility {
                 });
         }
 
+        public static getSavedPlugins(projectInfo: IProjectInfo): Q.Promise<Cordova.ICordovaPluginInfo[]> {
+            return TacoPackageLoader.lazyCordovaRequire(projectInfo.cordovaCliVersion)
```
MSLaguana: What if we are not in a taco project, so we don't have a cordovaCliVersion?

src/taco-cli/cli/kit.ts
digeff: It's another option.
Personally removesPluginsFromConfig makes me thing of a boolean
And removedPluginsFromConfig makes me thing of an array of plugins.
What do those names tell to you? If more people find the removed more intuitive, then I'll change it

src/taco-cli/cli/kit.ts
digeff: I added:
            // We remove the plugins that are overriden by the target kit from the config.xml
            // because the config.xml has preference over the kit, and if we don't do this we'll
            // break the kit. We need also to do this before changing the platforms, because changing
            // a platforms automatically adds plugins found on the config.xml file

src/taco-utils/cordovaHelper.ts
digeff: It seems to end up using cordova@ similar to doing:
npm install cordova@
And it seems to work.

src/taco-cli/cli/kit.ts
nisheetjain: I liked removedPluginsFromConfig more

PR #227, opened 2015-12-22
Creating Kits v1.1 and v2.2 that just pin "missing" plugins into the versions most suitable for the CLIs used

digeff:
Most of the PR is Just copying templates/default into templates/fixed-whitelist.


src/taco-kits/TacoKitMetadata.json
```diff
@@ -56,6 +57,150 @@
...
+                },
+                "cordova-plugin-legacy-whitelist": {
+                    "version": "1.1.0"
+                }
+
+            }
+        },
+        "tacokit@2.1.0": {
+            "cordova-cli": "5.2.0",
+            "releaseNotesUri": "http://go.microsoft.com/fwlink/?LinkId=623759",
```
MSLaguana: This release notes link will need to be updated surely? To mention that these aren't validated plugins, but point-in-time?

MSLaguana:
Were you also going to add a tacokit@3.0.0 for cordova 5.4.1, or are we waiting until cordova 6.0 or that?

digeff:
We are waiting...

src/taco-kits/TacoKitMetadata.json
digeff: Yes, we need to remember to do that!

meenakb: +1. Please open a bug for tracking it.

meenakb:
:+1:

PR #230, opened 2015-12-28
Add livereload support for iOS simulators

omefire:
iOS simulators can only connect to the 'local' endpoint, instead of the 'external' one.
This PR enables that scenario.


src/taco-livereload/lib/utils/homePage/bower_components/q/.npmignore
```diff
@@ -0,0 +1,23 @@
+node_modules
```
MSLaguana: Is this correct? We want to not package up the dependencies of the q package used by bower? Do we `bower install` at runtime?

MSLaguana: And should we be including the bower version of Q verbatim in our codebase? Wouldn't it be better to take a dependency on bower and acquire Q on demand, so we can take fixes without requiring a code change if possible?

omefire: we do not `bower install` at runtime, it's currently packaged with TACO.

omefire: I'll make changes to the code, so that it brings in Q at runtime.

nisheetjain: we shouldn't be committing installed packages?

nisheetjain: can you describe the scenario which requires bower components?

omefire: We do not need 'bower_components', it was just a way to acquire the dependencies.

omefire: @MSLaguana On second thoughts, I think we should package this code with the html code (it's standard practice). Otherwise, the end-user will need to be connected to the internet in order to use livereload (which is not a desirable constraint, IMO).

src/taco-livereload/lib/utils/Patcher.js
```diff
@@ -70,25 +77,27 @@ Patcher.prototype.removeCSP = function () {
...
+
+    // 1- copy the 'homePage' directory
+    var pluginHomePageDir = path.join(__dirname, 'homePage');
     
-    return HOME_PAGE + randomizedString;
+    // Append random string to 'homePage' directory name 
+    // ... so that it doesn't clash with potential user-defined 'homePage' folders 
+    var homePageDir = 'homePage_' + Math.random();
+    var homePageDirFullPath = path.join(projectRoot, multiPlatforms.getPlatformWWWFolder(platform), homePageDir);
+    return Q.nfcall(ncp, pluginHomePageDir, homePageDirFullPath).then(function() {
```
nisheetjain: you can use taco-utils.UtilHelper.copyRecursive instead

src/taco-livereload/lib/utils/homePage/homePage.html
```diff
@@ -0,0 +1,103 @@
...
+        var SERVERS = JSON.parse('__SERVERS__');
+        
+        // Note: These are the kinds of servers we are interested in. (e.g: we don't care about 'ui' and 'ui-external')
+        var serverTypes = ['local', 'external', 'tunnel'];
+        
+        var areServersAccessiblePromises = serverTypes.map(function(serverType) {
+            return isServerAccessible(SERVERS[serverType]);
+        });
+        
+        Q.all(areServersAccessiblePromises).then(function(areServersAccessible) {
```
nisheetjain: can we use Q.any here?

joshuaskelton: I don't believe we can. The code appears to build an array of booleans indicating if we can ping a sever. We then check those responses in order of preference. Does Q.any guarantee ordering of execution?

src/taco-livereload/lib/utils/Patcher.js
joshuaskelton: I've updated the code to use UtilHelper.copyRecursive

PR #238, opened 2016-01-08
V1.2.1

leotlee:
Merge back to master + post release version bump


src/taco-remote-multiplexer/dynamicDependencies.json
```diff
@@ -1,7 +1,7 @@
 ﻿{
   "latestTacoRemoteLib": {
     "packageName": "taco-remote-lib",
-    "packageId": "taco-remote-lib@2.0.2",
+    "packageId": "taco-remote-lib@2.0.3",
```
MSLaguana: Do you need to bump this too to match taco-remote-lib's bumped version?

leotlee: Good catch, thanks.

PR #244, opened 2016-01-13
Fix remote build log issues

guillaumejenkins:
-Fixed TACO red error messages interleaving with remote build output on the client console
-Fixed build server reusing build log from previous builds


nisheetjain:
:+1:

PR #247, opened 2016-01-14
Fixing logger.logXXX to handle undefined argument

nisheetjain:


nisheetjain:
@leotlee can you please review this?

src/taco-utils/test/logger.ts
```diff
@@ -0,0 +1,51 @@
...
+        verifyScenario("<h1>hello world</h1>");
+    });
+    it("should support known tags with strings", function (): void {
+        verifyScenario("<helptitle>hello world</helptitle>");
+    });
+    it("should support undefined strings", function (): void {
+        verifyScenario(undefined);
+    });
+    it("should support null strings", function (): void {
+        verifyScenario(null);
```
leotlee: Let's add:

NaN
empty string ("")
0
false

src/taco-utils/logger.ts
```diff
@@ -24,7 +24,9 @@ module TacoUtility {
          * if using any kind of formatting, make sure that it is well formatted
          */
         public static log(message: string): void {
```
leotlee: Just double-checking, does it work now with "remotebuild test"?

src/taco-utils/test/logger.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+});
+
+function verifyScenario(input: string): void {
+    try {
+        Logger.log(input);
+        Logger.logError(input);
+        Logger.logWarning(input);
+        should(true).be.ok;
```
guillaumejenkins: I think should.js has a way of saying shouldNotThrow({...code here})

guillaumejenkins: should(code_block).not.throw([error])

nisheetjain: I like it

src/taco-utils/test/logger.ts
nisheetjain: Logger.log takes only string as input. Added the empty string case. skipped others

leotlee: :+1:

PR #249, opened 2016-01-21
Updating package versions for taco-cli (feature addition, live-reload…

nisheetjain:
…) and taco-kits (new kits)


nisheetjain:
got this offline reviewed from @leotlee

PR #252, opened 2016-02-03
Set ios-sim to a known good version

joshuaskelton:
ios-sim has an issue with launching the simulator for a different devicetypeid when the simulator is already running. We are working around this by setting our dependency to a known good version.

ios-sim issue:
https://github.com/phonegap/ios-sim/issues/187


MSLaguana:
Looks good to me. Hopefully they'll get a fix in ios-sim soon so we can go back to taking the latest.

PR #257, opened 2016-02-05
Fix for issue #256

### LINKED ISSUE #256: Can't run iOS device using cordova-ios@4.X
When using cordova-ios@4.0.1 to run on a device, it tries to run a slice on `runOptions.argv` which we are passing in as null. This particular issue can be fixed by us passing in an empty array, but there may be other changes to the API that we haven't caught yet so we'll need to check that over.

guillaumejenkins:
Changed Cordova raw options conversion for 5.4.0+ api: argv field now defaults to empty array instead of null

Fixes #256


MSLaguana:
Looks good to me

nisheetjain:
:+1:

Issue #258, opened 2016-02-10
Use common npm sub-module for the utility functions needed for browsersync on cordova 


asselin:
I recently factored out the utility functions needed to enable browersync on cordova into a separate github repo and npm module (https://github.com/PointSource/cordova-browsersync-primitives).  These are the functions to add CSP, fix ATS, etc.  I did this because I need these functions for a project I'm working on, and I know there are also 2 existing projects that also use them (https://github.com/nparashuram/cordova-plugin-browsersync and https://github.com/Microsoft/TACO).

The purpose of this issue is to suggest that we consolidate around 1 implementation of these utility functions that we can all pull in as an npm submodule.

What's your opinion of converting over to use a common implementation?

guillaumejenkins:
Thanks for the suggestion. We think it's a good idea, but unfortunately at the moment we don't have the resources to work on this. We will gladly accept PRs, though!

nisheetjain:
hey @asselin isn't https://www.npmjs.com/package/taco-livereload what you are looking for? if you want to really expose sub-components like addCSP ATSRemover, then that can be done very cheaply. As @guillaumejenkins said, feel free to send a PR. :)

asselin:
@nisheetjain Right, what I did was to expose the subcomponents so that we can share them across the 3 projects as a separate npm module.  I'll work on a PR.

cc: @guillaumejenkins

asselin:
Started work on this under the https://github.com/PointSource/TACO/tree/feature/refactor_primitives branch.  Not ready for a PR just yet, but wanted to update status.

asselin:
I finished the work on a PR for this about a week ago.  Is there anything blocking merging in the PR?

axemclion:
@asselin, sorry for the delay in response. I spoke to the team, and looks like we may not be able to merge this in. We are exploring an alternate way to do livereload where we no longer have issues with protocols, and we may not be consuming browsersync as the new module.

PR #263, opened 2016-02-17
Switch to cordova-browsersync-primitives

asselin:
This is to address issue #258 .


src/taco-livereload/lib/utils/Patcher.js
```diff
@@ -31,63 +28,50 @@ function Patcher(projectRoot, platforms) {
...
-        return cspRemover.Remove().then(function () {
-            var atsRemover = new ATSRemover(self.projectRoot, plat);
-            return atsRemover.Remove();
-        }).then(function () {
-            var platformIndexUrl = url.resolve(serverUrl, path.join(multiPlatforms.getPlatformWWWFolder(plat), self.startPage));
-            return copyHomePage(self.projectRoot, plat, platformIndexUrl);
-        }).then(function (homePage) {
-            return helpers.ChangeStartPage(self.projectRoot, plat, homePage); 
-        });
+        if (plat == 'ios')
```
MSLaguana: This should use `===`, we have a `gulp tslint` task that enforces style constraints like this.

MSLaguana: Oh I see that this is actually a js file and so the tslint wouldn't pick it up anyway. Still good to try and be consistent I think

asselin:
@MSLaguana np, I went ahead and made the change from == to ===

MSLaguana:
The PR looks good to me now, however we can't accept it until you fill out the CLA as mentioned by the bot earlier on.

src/taco-livereload/lib/utils/Patcher.js
```diff
@@ -31,63 +28,50 @@ function Patcher(projectRoot, platforms) {
...
-    return platformWwws.forEach(function (platWWWDir) {
-        var platformIndexLocal = path.join(platWWWDir, startPage);
-        var remover = new CSPRemover(platformIndexLocal);
-        remover.Remove();
-    });
-};
+    this.platforms.forEach(function(plat) {
+        var platWWWFolder = browserSyncPrimitives.getWWWFolder(plat);
+        var platformIndexLocal = path.join(this.projectRoot, platWWWFolder, this.startPage);
+        browserSyncPrimitives.addCSP(platformIndexLocal);
```
guillaumejenkins: Is the call to `addCSP` intentional? From the method's name (`removeCSP`), and the previous implementation (`new CSPRemover(platformIndexLocal).Remove();`) I get the feeling this might be a mistake?

asselin: @guillaumejenkins Yes, it's intentional.  The confusion comes from conceptually what is going on here.  It's ADDing text to the CSP line to REMOVE some restrictions.  If you find this confusing, I can rename everything to a more neutral API name, such as updateCSPTag, for example.

guillaumejenkins: Gotcha; I think it's fine as it is

joshuaskelton:
On our side of things, what impact does this have about adding an additional dependency? 

@asselin looking at your repo I don't see a readme or license file. The npm page states that it is licensed under Apache 2.0. Can you please clarify?

asselin:
Yes, Apache.  Lack of README and LICENSE was an oversight on my part. I'll push an update today to add those files.

asselin:
Repo updated and published to npm.

asselin:
@joshuaskelton Are you waiting on anything from me to merge this PR?

axemclion:
I think we should not be merging this. We are exploring a different way to do live reload - where we don't switch the protocol of the webview (which leads into issues) and that would not consume browser-sync as the module exposes. 

@guillaumejenkins @joshuaskelton, My recommendation would be not to merge this PR, at least till we look at the feasibility of the new approach.

MSLaguana:
Sorry for the extended delay. We now suggest that people use our [VSCode extension](https://marketplace.visualstudio.com/items?itemName=vsmobile.cordova-tools) for livereload support, either via ionic or simulating in a browser.

PR #264, opened 2016-02-18
Fix for #259: Always prepare remote platforms

### LINKED ISSUE #259: Plugins not working on full build, fine after incremental build
Cordova App, deployed from Visual Studio 2015 to a Mac build server. Cordova 5.4.1
The first build always fails. The ondeviceReady() function is called, but all plugins like window.StatusBar are null.
When I then deploy again, an incremental build is generated, and the error is gone.

This happens every time. Is there anything I can check?

guillaumejenkins:
The newer Cordova versions have stopped implicitly calling "prepare" when a platform is added. We must now always manually prepare platforms, whether we add the platform or not.

Fixes #259.

### DIFF
--- a/src/taco-remote-lib/common/builder.ts
+++ b/src/taco-remote-lib/common/builder.ts
@@ -139,7 +139,6 @@ class Builder {
 
     /**
      * Adds the platform if it isn't already present, or removes and re-adds it if the project Cordova version changed, or config.xml has a saved platform version that differs from the installed one.
-     * Prepares the platform afterwards.
      */
     private ensurePlatformAdded(): Q.Promise<any> {
         if (!fs.existsSync(path.join("platforms", this.currentBuild.buildPlatform))) {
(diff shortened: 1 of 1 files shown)


src/taco-remote-lib/common/builder.ts
```diff
@@ -162,7 +153,7 @@ class Builder {
         if (xmlVersion) {
             mustReAddPromise = this.getCurrentPlatformVersion().then((currentVersion: string) => {
                 if (!semver.satisfies(currentVersion, xmlVersion)) {
```
MSLaguana: It looks like this section is redundant now since we always invoke prepare rather than conditionally

MSLaguana: Oh nevermind This is for re-adding the platform, not whether we need to prepare. Disregard.

src/taco-remote-lib/common/builder.ts
```diff
@@ -132,28 +132,19 @@ class Builder {
             fs.mkdirSync("platforms");
```
MSLaguana: Might be worth changing this function name to `addAndPreparePlatform` to be more accurate.

MSLaguana:
Looks good to me

nisheetjain:
:+1:

PR #267, opened 2016-02-24
Fixed compile errors due to TypeScript 1.8.2

guillaumejenkins:


src/taco-remote/lib/requestRedirector.ts
```diff
@@ -35,13 +35,9 @@ class RequestRedirector implements TacoRemoteLib.IRequestRedirector {
                 var packageSpec: IPackageSpec = mux.getPackageSpecForQuery(req.query);
                 return TacoPackageLoader.lazyTacoRequire<IRemoteLib>(packageSpec.packageKey, packageSpec.dependencyConfigPath);
             })
-            .catch(function (err: any): IRemoteLib {
+            .catch(function (err: any): IRemoteLib | any {
```
MSLaguana: Is that `any` actually required? I would hope not...

src/taco-tests-utils/mockCordova.ts
```diff
@@ -31,7 +31,7 @@ module TacoTestsUtils {
             }
         }
 
-        function notImplemented<T>(): T {
+        function notImplemented<T>(): T | any {
```
digeff: Why is the any needed?

PR #271, opened 2016-03-02
When attempting to debug a device, we now check whether the web inspe…

MSLaguana:
…ctor is enabled


src/taco-remote-lib/ios/ios.ts
```diff
@@ -266,12 +266,32 @@ class IOSAgent implements ITargetPlatform {
...
+                if (dataStr.match(/Unable to connect to/)) {
+		    var error = new Error(resources.getStringForLanguage(req, "WebInspectorDisabled"));
+		    deferred.reject(error);
+		}
+		if (dataStr.match(/Connected :[0-9]+ to/)) {
+		    deferred.resolve();
+		}
+            });
+	    // Allow time to discover devices to register.
+	    Q.delay(10000).then(() => deferred.reject(new Error(resources.getStringForLanguage(req, "NoDevicesFound"))));
```
digeff: Are we sure 10 secs is enough for this? Even in slow machines?

digeff:
It looks good apart from the minor question

src/taco-remote-lib/ios/ios.ts
MSLaguana: When I timed it myself it was taking about 750ms. I believe that VS currently times out after 30s (it uses a common timeout for launching and debugging)

PR #279, opened 2016-03-17
Separate ios launch

MSLaguana:
Switch to using the now-published-separately idevice-app-launcher package for launching iOS apps. This includes a fix that means we properly stop the app if it runs into native crashes rather than keeping the phone stopped in the debugger, and we also now bubble up those crashes to the BuildInfo so be reported to clients if they ask.


src/taco-remote-lib/ios/ios.ts
```diff
@@ -80,23 +83,30 @@ class IOSAgent implements ITargetPlatform {
...
+                    if (isCrash) {
+                        buildInfo.updateStatus(BuildInfo.CRASHED);
+                    } else {
+                        buildInfo.updateStatus(BuildInfo.TERMINATED);
+                    }
+                });
+            }).then(function(success: net.Socket): void {
+                res.status(200).send(buildInfo.localize(req, resources));
+            }, function(failure: any): void {
+                if (failure.message) {
```
dlebu: All this failure message processing can be extracted in a different method. Would we be able to reuse it for any other cases?

src/taco-remote-lib/package.json
```diff
@@ -28,7 +28,8 @@
         "plist-with-patches": "0.5.1",
         "taco-utils": "file:../taco-utils",
         "semver": "^4.3.6",
-        "archiver": "^0.16.0"
+        "archiver": "^0.16.0",
+	"idevice-app-launcher": "^0.1.0"
```
dlebu: nit: whitespace formatting

dlebu:
Minor comments. LGTM!

src/typings/idevice-app-launcher.d.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+/// <reference path="./node.d.ts" />
+/// <reference path="./q.d.ts" />
+
+declare module "idevice-app-launcher" {
```
meenakb: How about including this typing file in https://github.com/Microsoft/idevice-app-launcher ?

meenakb:
:+1:

PR #280, opened 2016-03-21
Updating TacoPackageLoader

MSLaguana:
With this change we to back to invoking npm on the command line rather than via its unstable internal APIs. Also fixing an issue where valid github URLs with a #branch specifier were misunderstood.

### DIFF
--- a/src/taco-utils/npmHelper.ts
+++ b/src/taco-utils/npmHelper.ts
@@ -7,76 +7,69 @@
 
 "use strict";
 
-import npm = require("npm");
+import child_process = require("child_process");
+import os = require("os");
 import path = require("path");
 import Q = require("q");
-import nopt = require("nopt");
-import url = require("url");
-import semver = require("semver");
-
-var npmconf = require("npm/lib/config/core.js");
-var configDefs = npmconf.defs;
-var types = configDefs.types;
-var shorthands = configDefs.shorthands;
 
 import installLogLevel = require("./installLogLevel");
 import InstallLogLevel = installLogLevel.InstallLogLevel;
 
 module TacoUtility {
     export class NpmHelper {
-        private static runNpmCommand(npmCommand: string, args: string[], workingDirectory: string, commandFlags?: string[], logLevel?: InstallLogLevel, optionalArgs?: any): Q.Promise<any> {
-            var conf: any = {};
-            if (commandFlags) {
-                conf = nopt(types, shorthands, commandFlags, 0);
-            }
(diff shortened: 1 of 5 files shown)


src/taco-utils/npmHelper.ts
```diff
@@ -7,76 +7,69 @@
...
-            var oldValues: any = {};
-            return Q.nfcall(npm.load, {}).then(function() {
-                var configList = npm.config.list[0];
+            const npmExecutableName = "npm" + (os.platform() === "win32" ? ".cmd" : "");
 
-                propertiesToSet.forEach(prop => {
-                    oldValues[prop] = configList[prop];
-                    configList[prop] = conf[prop];
-                });
+            const npmProcess = child_process.spawn(npmExecutableName, spawnArgs, options);
```
meenakb: Why are we going back to using npm command? Is there a specific stability issue with using the npm module?

src/taco-utils/npmHelper.ts
```diff
@@ -7,76 +7,69 @@
...
-                    } else {
-                        delete configList[prop];
-                    }
+                npmProcess.stderr.on("data", (data: Buffer) => {
+                    stderr += data.toString();
                 });
+            }
+
+            npmProcess.on("exit", (code: number, signal: string) => {
+                if (code === 0) {
```
meenakb: Should we reject the promise if the command exits with a non-zero error code?

src/taco-utils/tacoPackageLoader.ts
```diff
@@ -96,7 +96,7 @@ module TacoUtility {
         // Need to declare FILE_REGEX_PREFIX before FILE_URI_REGEX, but don't want to expose it outside the class
         private static FILE_REGEX_PREFIX: string = "file://";
         private static CORDOVA_NPM_PACKAGE_NAME: string = "cordova";
-        public static GIT_URI_REGEX: RegExp = /^http(s?)\\:\/\/.*|.*\.git$/;
+        public static GIT_URI_REGEX: RegExp = /(^http(s?):\/\/.*)|(.*\.git(#.*)?$)/;
```
meenakb: Good catch!

src/taco-utils/npmHelper.ts
MSLaguana: A couple of reasons. The one specific thing that prompted this was when I tried to install cordova from a git repo, I got errors because the npm code assumed that "The current process" is the npm executable, which meant it tried to run scripts using "the current process" which might actually be remotebuild for example. Npm really doesn't want to be used via require, only as an executable.

src/taco-utils/npmHelper.ts
MSLaguana: Yes we should, good catch there.

meenakb:
:+1:

PR #284, opened 2016-03-31
Changing taco-remote-lib to use cordova-plugin-vs-taco-support

MSLaguana:
To resolve issues with before/after build hooks not running, this commit removes some of the workarounds for outdated problems, and changes the implementation of res/native functionality to use the cordova-plugin-vs-taco-support.

### DIFF
--- a/src/taco-remote-lib/common/builder.ts
+++ b/src/taco-remote-lib/common/builder.ts
@@ -60,7 +60,7 @@ class Builder {
             .then(function (): void { self.currentBuild.updateStatus(BuildInfo.BUILDING, "UpdatingPlatform", self.currentBuild.buildPlatform); process.send(self.currentBuild); })
             .then(function (): Q.Promise<any> { return self.beforePrepare(); })
             .then(function (): Q.Promise<any> { return self.addPlatform(); })
-            .then(function (): Q.Promise<any> { return self.build_platform(); })
+            .then(function (): Q.Promise<any> { return self.buildPlatform(); })
             .then(function (): Q.Promise<any> { return self.afterCompile(); })
             .then(function (): void { self.currentBuild.updateStatus(BuildInfo.BUILDING, "PackagingNativeApp"); process.send(self.currentBuild); })
             .then(function (): Q.Promise<any> { return isDeviceBuild ? self.package() : Q({}); })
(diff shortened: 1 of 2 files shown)


MSLaguana:
@chuxel Do you see anything that I've missed / important differences from the vsts cordova implementation?

Chuxel:
One caution is that Cordova 4.3.1 does not support npm based plugin installs.  In taco-team-build we fallback to the Git URI if Cordova < 5.0.0

MSLaguana:
Cordova < 5.0.0 will fail to acquire any new plugins that aren't from git / file paths since the plugin repository was EOL'd... That doesn't impact this specific change too much since this taco-remote-lib package is only used for cordova >= 5.4.0, but when backporting it to deal with earlier cordovas I'll have to keep that in mind it's true, just in case someone is still using such an old version and happens to have the plugins installed already.

Chuxel:
Ah, I hadn't realized this is only used for new versions - is the plan to backport or no?

Anyway, yes, the old plugin repository is read only, so old plugin IDs work, but you can't add any new ones. It is "deprecated" but not shut down yet. So, to support Cordova 4.3.1, so for backwards compat you'd just add the support plugin via Git URI (unless Visual Studio will no longer support Cordova 4). taco-team-build does just that.

Chuxel:
Oh 6.0.0 did drop support for the old plugin repo, but that does not apply to earlier versions of Cordova.

MSLaguana:
Ah I had misunderstood the deprecation, I thought that the old plugin repo had been taken down and not just made read-only. In that case when I backport it I'll make sure to add via git URL.

MSLaguana:
You can check out the relevant changes for pre 5.4.0 at https://github.com/Microsoft/TACO/compare/817c...pre540TacoPlugin

src/taco-remote-lib/common/builder.ts
```diff
@@ -277,33 +249,15 @@ class Builder {
...
-            var resTo: string = path.join("platforms", this.currentBuild.buildPlatform);
-            return UtilHelper.copyRecursive(resFrom, resTo);
-        }
-
-        return Q({});
-    }
-
-    private compile_platform(): Q.Promise<any> {
-        Logger.log("cordova compile " + this.currentBuild.buildPlatform);
+    private build_platform(): Q.Promise<any> {
```
joshuaskelton: camelCase?

PR #308, opened 2016-08-30
Filtered out carriage return from CordovaWrapper.getGlobalCordovaVersion

gregvis:
vcordova was returning "6.3.1\r" for me which was throwing errors at the remotebuild. I've set up the getGlobalCordovaVersion to filter the "\r".


MSLaguana:
We have deprecated the taco-cli now. The sources have moved to http://github.com/microsoft/taco-cli but we will no longer be maintaining it.

